PR #12, opened 2022-07-01
Cleanup + Implement Read

harribo11:


examples/schema-test/main.go
```diff
@@ -34,13 +124,14 @@ func (s *Strct) Annotate(a resource.Annotator) {
 func main() {
 	println(reflect.TypeOf((*Enum)(nil)).Elem().String())
```
iwahbe, line 125: Why do we do this? I think it will invalidate the provider, since we assume that providers print their address first?

resource/resource.go
```diff
@@ -107,7 +107,7 @@ type CheckFailure struct {
 
 type Read interface {
 	Custom
-	Read(ctx Context) error
+	Read(ctx Context, id string) (*pulumirpc.ReadResponse, error)
```
iwahbe, line 110: I agree that we should change the method to accept the `id`, but the user **should not** need to return a `pulumirpc.ReadResponse`. The library should fabricate a `ReadResponse` by examining the assigned fields of the resource the method was called on, just like `Update` or `Create`.

internal/inputmap/inputmap.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type InputToImplementor map[reflect.Type]reflect.Type
+
+func (m InputToImplementor) add(k interface{}, v interface{}) {
+	m[reflect.TypeOf(k).Elem()] = reflect.TypeOf(v).Elem()
+}
+
+func GetInputMap() InputToImplementor {
```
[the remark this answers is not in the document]
iwahbe: Since we aren't really getting an existing map, but creating a new one, why not call it `New`?

schema.go
```diff
@@ -44,46 +44,23 @@ type serializationInfo struct {
...
-	spec.Name = opts.Name
-	spec.Version = opts.Version.String()
-
 	info := serializationInfo{}
 	info.pkgname = opts.Name
 	info.resources = make(map[reflect.Type]string)
 	info.types = make(map[reflect.Type]string)
 	info.enums = make(map[reflect.Type]string)
-	info.inputMap = initializeInputMap()
+	info.inputMap = inputmap.GetInputMap()
```
iwahbe, line 57: It's idiomatic to construct this in place, instead of creating the zero initialized struct and adding member fields.

schema.go
```diff
@@ -769,20 +546,20 @@ func getTypeKind(t reflect.Type) (string, bool) {
 	return typeName, isEnum
 }
 
-func serializeTypeRef(t reflect.Type, info serializationInfo) (*schema.TypeSpec, error) {
+func (info serializationInfo) serializeTypeAny(t reflect.Type) (*schema.TypeSpec, error) {
 	typeKind, enum := getTypeKind(t)
 	if enum {
-		enumSpec, err := serializeEnum(t, info)
+		enumSpec, err := info.serializeTypeEnum(t)
```
iwahbe, line 552: `serializeEnumType` is easier to read then `serializeTypeEnum`. Especially since other versions of this function put the type kind in the middle, see `serializeReferenceType`.

schema.go
```diff
@@ -866,21 +643,10 @@ func serializeType(typ any, info serializationInfo) (schema.ComplexTypeSpec, err
...
-			Type:        typeKind,
-			Description: descriptions[""],
-		},
-		Enum: enumVals,
-	}, nil
+	return schema.ComplexTypeSpec{}, fmt.Errorf("Types must be structs - use provider.Enum to create an enum type %s", t)
 }
 
-func serializeEnum(t reflect.Type, info serializationInfo) (schema.TypeSpec, error) {
+func (info serializationInfo) serializeTypeEnum(t reflect.Type) (schema.TypeSpec, error) {
```
iwahbe, line 649:
```suggestion
func (info serializationInfo) serializeEnumType(t reflect.Type) (schema.TypeSpec, error) {
```

schema.go
```diff
@@ -942,9 +708,21 @@ func mergeMapsOverride[T any](base, override map[string]T) map[string]T {
 	return base
 }
 
+func baseType(i any) reflect.Type {
```
iwahbe, line 711: What is the benefit of this function? Does a `baseType(i)` have any properties that we care about?

schema.go
harribo11: There is already a function named `serializeEnumType` which returns a `schema.ComplexTypeSpec`. Maybe we could rename all functions which return `schema.TypeSpec` to `serializeAny`, `serializeReference`, `serializeEnum`?

examples/schema-test/main.go
harribo11: Ah, meant to delete that. Also I wasn't aware of that.

schema.go
```diff
@@ -590,23 +368,24 @@ func serializeResource(rawResource any, info serializationInfo) (schema.Resource
...
-		isInput := field.Type.Implements(reflect.TypeOf(new(pulumi.Input)).Elem())
-		_, isOutput := vField.Interface().(pulumi.Output)
+		isInputType := field.Type.Implements(reflect.TypeOf(new(pulumi.Input)).Elem())
+		_, isOutputType := vField.Interface().(pulumi.Output)
 
 		for fieldType.Kind() == reflect.Ptr {
 			fieldType = fieldType.Elem()
 		}
-		if isOutput {
+		//Need to fetch the underlying types for input and outputs
```
iwahbe, line 377: Nit: Comments should have space after the `//`. This applies in multiple places.

iwahbe (approved):
A couple of small things to fix, but it looks good to merge otherwise.

PR #17, opened 2022-07-07
Attempt to port a serverless component to this library

lukehoban:
This is the beginnings of an attempt to use this library to use this library to port a slightly non-trivial component.  A few questions/issues that came up - curious if there are thoughts on these, or features that need to be added to support these:
1. Ran into https://github.com/iwahbe/pulumi-go-provider/issues/16 trying to use external types in outputs (but was able to remove this temporarily to unblock further issues).
2. The use of a locally defined FunctionMapInput is not supported - it seems like only Input/Output wrappers over built-in `pulumi` SDK types are allowed.  How are nested types supported?  I see in the provider implementations that plain types are used, so these aren't needed.  But in the components, they seem to (rightly?) expect to work with inputs/outputs that are Inputs/Outputs.  But then how do I use nested types within this?
3. Related - note that I have to copy the same giant `types.go` file used in https://github.com/pulumi/pulumi-package-serverless/ (which I am porting from).  Not everything there is strictly necessary - but a lot of it seems to be.  Is there a better way to author this with this model?  Note that in the original variant of this, I actually wrote schema first, then generated SDK, then stole the SDK types back into the implementation to use for the implementation of the Construct.  Convoluted for sure - but with this amount of nested types - authoring in schema was actually quite a lot simpler than in Go types.  Can this be easier with the higher level framework here?
4. Not yet directly related to this - but would be *lovely* to have testing support built into the examples - a `main_test.go` that runs ProgramTest with a YAML program (so that SDK isn't needed for inner loop testing)??

Currently this is blocked on (2) above - resulting in:

```
Error: input Functions for property main.Service has type main.FunctionMapInput, which is not a valid input type
```


lukehoban:
cc @iwahbe @harribo11 for thoughts on questions/notes above.

harribo11:
Concur with Ian on points 1 and 2. Will our token system work across all providers? For example the docker provider has type tokens formatted `docker:index/container:Container`. If not then we will probably need to add some way to handle those edge cases.

examples/serverless/main.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func (s *Service) Construct(name string, ctx *pulumi.Context) error {
+	component := &Service{}
+	// What about opts?
+	err := ctx.RegisterComponentResource("serverless:index:Service", name, component)
+	if err != nil {
+		return err
+	}
+
```
iwahbe, line 68:
```suggestion
```
`s` is the component that you are registering here. You don't need to create a new component or register it. We don't expose Options, since they should all be inherited by parenting `s`. Let me know if you can think of a reason we need to expose `Options`.

examples/serverless/main.go
```diff
@@ -0,0 +1,247 @@
...
+	// TODO[pulumi/pulumi#6073]: Workaround issue with not being able to directly return outputs from Apply.
+	component.Url = out.ApplyT(func(v interface{}) *string {
+		return v.(*string)
+	}).(pulumi.StringPtrOutput)
+
+	if err := ctx.RegisterResourceOutputs(component, pulumi.Map{
+		"url": component.Url,
+	}); err != nil {
+		return err
+	}
```
iwahbe, line 218:
```suggestion
	s.Url = out.ApplyT(func(v interface{}) *string {
		return v.(*string)
	}).(pulumi.StringPtrOutput)
```

iwahbe:
Closing, since this is out of date.

PR #21, opened 2022-07-07
Added input type detection using reflection

harribo11:


iwahbe (commented):
The logic looks great. Would it be possible to 
1. Refactor out a function: `underlyingType(typ: reflect.Type) -> (reflect.Type, error)` that takes the field type and returns its underlying type, doing the correct behavior regardless of if the underlying type is an input, output or normal type.
2. Add some unit tests for the new `underlyingType` function. We should test against both internal and external types.

lukehoban:
Btw - in case it’s helpful - there’s some very similar conversion logic in the component provider mapping interface here: https://github.com/pulumi/pulumi/blob/664ed0ec3983c50ddd43655375966c11c013959e/sdk/go/pulumi/provider.go#L461

schema_test.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"reflect"
+	"testing"
+
+	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/ec2"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnderlyingType(t *testing.T) {
```
iwahbe, line 26: I think this would really benefit from a technique called table based testing. We do this all over, but a good example can be found here: https://github.com/pulumi/pulumi/blob/7827eb26bc1a64aa56ad450d35bccde0ceadd78e/nil#L10-L51

iwahbe (approved):

PR #24, opened 2022-07-10
Add invokes

### LINKED ISSUE #22: Missing Support for Invokes

iwahbe:
Fixes #22 

## Design decisions

`provider.Functions` is a new top level option when constructing a provider. It takes a `function.Function` which is a wrapper around a go `func`. Idiomatic construction is to use `function.New(f, "description")`. 

### Valid Functions
Unfortunately, it is not possible to recover runtime information about argument names, so this restricts what we can infer about the `Input` and `Output` types for a function. Instead, we mandate that input and output types are explicit types. They obey the same rules and restrictions as elements in the `provider.Types` option. To allow good practice, functions can optionally accept a `context.Context` in the first position (bumping their argument to the second) and may return an error. This makes the full signature of a valid function:
```
func([context.Context] [InputT]) ([OutputT] [error])
```

Examples can be found in the `str` example provider.


examples/str/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+func main() {
+	err := p.Run("str", semver.Version{Minor: 1},
+		p.Functions(
+			function.New(Replace,
+				"Replace returns a copy of the string s with all\n"+
+					"non-overlapping instances of old replaced by new.\n"+
+					"If old is empty, it matches at the beginning of the string\n"+
+					"and after each UTF-8 sequence, yielding up to k+1 replacements\n"+
+					"for a k-rune string."),
```
harribo11, line 23: It feels potentially confusing that for resources you have to use annotation but here the description is just outright specified.

internal/introspect/invokes.go
```diff
@@ -0,0 +1,70 @@
...
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+// Retrieve the Input type of a function, if any.
+func InvokeInput(f reflect.Type) (input reflect.Type, hasContext bool, err error) {
+	contract.Assert(f.Kind() == reflect.Func)
+	isContext := func(t reflect.Type) bool { return t.Implements(reflect.TypeOf(new(context.Context)).Elem()) }
+	badTypeMsg := fmt.Errorf("Functions must be of the type func([context.Context, ], T). Found type %s", f.String())
+	switch f.NumIn() {
```
harribo11, line 30: I don't understand why we limit the signature of a function so strictly. I see that most/all of our functions on the registry have some args struct as input and a result struct as an output, could we not automatically wrap the the inputs and outputs into a struct?

internal/introspect/invokes.go
```diff
@@ -0,0 +1,70 @@
...
+			return nil, false, badTypeMsg
+		}
+		return f.In(1), true, nil
+	default:
+		return nil, false, badTypeMsg
+	}
+}
+
+// Retrieve the Output type of a function, if any.
+func InvokeOutput(f reflect.Type) (output reflect.Type, hasError bool, err error) {
```
harribo11, line 50: Would something like "canError" be better here?

examples/str/main.go
iwahbe: I prefer the `Annotate` based method, since it allows the description to be close to the resource definition, but you can't implement methods on a `func`. I didn't want to mandate wrapping your function in a type and then implementing `Annotate` on that new type, especially since we only annotate the function, not its fields. I agree that it introduces a split in design. If you have a more intuitive design, I would love to hear it. I couldn't find a design that I loved here.

iwahbe: I also struggled to come up with a type for `provider.Functions`. This simplifies the problem, since every type is `function.Function`.

internal/introspect/invokes.go
iwahbe: The schema (and the gRPC protocol) have functions pass input structs and return output structs, so we need to end up with that. We can't do automatic wrapping of multiple inputs and outputs into their struct components because reflection does not provide us with argument names, so we won't know what keys to provide in the newly created structs. 

Given:
```go
func AddWithRemainder(x, y int) (sum int, remainder int) {
    return (int)((u8)x+y), (x+y)%(2^8) 
}
```
We can only retrieve
```go
reflect.TypeOf(AddWithRemainder).String() == "func(int, int) (int, int)"
```

internal/introspect/invokes.go
iwahbe: Probably.

PR #26, opened 2022-07-13
Input/Output Types For Component Prototype

### LINKED ISSUE #18: Ergonomic way to reference provider defined types in component resource
Right now, component resources use `Input` and `Output` types as their fields. This requirers naming the types in the struct. This is fine for types external to the pkg, but works poorly when constructing references the the SDK version of types defined in the provider. 

We need a more ergonomic solution to this problem.

See https://github.com/iwahbe/pulumi-go-provider/pull/17#issue-1296866304 for motivation.

harribo11:
This PR is a prototype for what a better way to reference inputs and outputs could look like. This should resolve #18 once the design is more polished.


types/generic.go
```diff
@@ -0,0 +1,92 @@
...
+func (in Input[T]) ToOutput() Output[T] {
+	return pulumi.ToOutput(in).(Output[T])
+}
+
+func (in Input[T]) ToOutputWithContext(ctx context.Context) Output[T] {
+	return pulumi.ToOutputWithContext(ctx, in).(Output[T])
+}
+
+func (in Input[T]) GetNilType() T {
+	return *(*T)(nil) //Hacky but needed for reflection
```
iwahbe, line 56:
```suggestion
var t T
return T
```

examples/hello-world/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+type HelloWorld struct {
+	r.Component
+	hInput types.Input[Hello] `pulumi:"hello" provider:"input"`
+	wInput types.Input[World] `pulumi:"world" provider:"input"`
+
+	helloworld types.Output[string] `pulumi:"helloworld" provider:"output"`
+}
+
+func (hw *HelloWorld) Create(ctx r.Context, name string, preview bool) (r.ID, error) {
```
iwahbe, line 58: I think this should be called `Construct`.

types/generic.go
```diff
@@ -0,0 +1,92 @@
...
+
+import (
+	"context"
+	"reflect"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+// GenericInput is an input type that accepts Generic and Output values.
+type GenericInput[T any] interface {
```
iwahbe, line 25: Can we turn this into just `Input` and hide the the `Input` struct as an implementation detail (`input`)?

iwahbe (commented):
I'm concerned with serialization. I don't know how Pulumi's struct serializers will handle `Input[T]`. 

If we can get this working, it would be amazing!

PR #41, opened 2022-08-16
Add cancel middleware

### LINKED ISSUE #30: Middleware: Cancel Propagation
Simpler use cases (like packages using `infer`) need to have timeouts hooked up to their `p.Context` arguments. To keep this step self-contained and orthogonal to other concerns, we should implement this as a middleware layer.

iwahbe:
This PR adds a new middleware that ensures passed in contexts are canceled as appropriate. 

There are two distinct problems this solves:
1. Applying timeouts to their appropriate contexts.
2. Applying the `Cancel` RPC request to all other contexts.

Fixes #30

### DIFF
--- a/middleware/cancel/cancel.go
+++ b/middleware/cancel/cancel.go
@@ -174,10 +174,11 @@ type entry[T any] struct {
 func (h *inOutCache[T]) insert(t T) (evict func() (missing bool)) {
 	h.m.Lock()
 	defer h.m.Unlock()
-	var i int
+	var i int // The index in values of the new entry.
 	if len(h.tombstones) == 0 {
-		i = len(h.values)
+		i = len(h.values) // We extend values.
 	} else {
+		// There is an empty slot in values, so use that.
 		i = h.tombstones[len(h.tombstones)-1]
 		h.tombstones = h.tombstones[:len(h.tombstones)-1]
 	}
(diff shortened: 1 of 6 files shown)


middleware/cancel/cancel.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+// Remove all values from the inOutCache.
+func (h *inOutCache[T]) drain() []T {
+	h.m.Lock()
+	defer h.m.Unlock()
+	h.inDrain = true
+	defer func() { h.inDrain = false }()
+	values := []T{}
+	// We reset tombstones.
```
aq17, line 219: Could the old cache `h` values just be returned along with a fresh cache, i.e. `return h.values, &inOutCache[T]`?

iwahbe: No, because `h.values` is of type `[]*entry[T]`. Some of `h.values` will be `nil`. We also need to call `evict` on all values we remove to indicate that they have already been handled.

I don't see the benefit of returning a new cache. Since a cache is inherently mutable, there is no reason to work towards immutability. We have already allocated the space for `h.values`, so we might as well use it. 

Can you explain why this is desirable?

aq17: That makes sense, I was just wondering if we wanted to have an upper bound on the size of the cache, or possibly reset it each time it was drained.

aq17 (approved):

Issue #38, opened 2022-08-15
Port a Native Provider: pulumi-command


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Create a full port of https://github.com/pulumi/pulumi-command, fully test it, and then replace the original implementation of pulumi-command with the port.

iwahbe:
The associated draft PR is https://github.com/pulumi/pulumi-command/pull/118.

PR #43, opened 2022-08-17
Bug fixes for command

iwahbe:
Part of #38 

Various bug fixes for problems I found porting over the command provider:
- The command provider makes extensive use of embedded fields. We should support this use case: https://github.com/pulumi/pulumi-go-provider/commit/41c622504e90eb57ce406b2ba4f199adf0047f46
- To avoid writing manual diffs whenever possible, we should enforce the `provider:"replaceOnChanges"` during an inferred diff: https://github.com/pulumi/pulumi-go-provider/commit/6fa2ef0766416108e5da04503f9b54f5b5089595
- We need to support embedding `resource.Asset` and `resource.Archive` in serialized structs, both at the type and schema level: https://github.com/pulumi/pulumi-go-provider/commit/7c73385922f962e37d15c5b2a9a15399afbb471e.

### DIFF
--- a/infer/configuration.go
+++ b/infer/configuration.go
@@ -133,7 +133,7 @@ func (c *config[T]) diffConfig(ctx p.Context, req p.DiffRequest) (p.DiffResponse
 	if c.t == nil {
 		c.t = new(T)
 	}
-	return diff[T, T, T](ctx, req, c.t, true)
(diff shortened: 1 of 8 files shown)


aq17 (approved):
LGTM

PR #44, opened 2022-08-18
Fix context and add default deps

iwahbe:
Part of #38 
- The first commit prevents nested contexts from unbounded recursion.
- The second defaults resources to mark all outputs as unknown if any inputs are unknown (and inputs are not explicitly wired to outputs).
- The last PR adds a default READ implementation so that refresh works out of the box.


iwahbe:
To speed up development time (and since this fixes some bad bugs), I'm merging without approval. I need this to unblock tests on #38.

PR #45, opened 2022-08-19
Handle nested secrets + Concurrency hardening

iwahbe:
This PR changes default secret handling to accommodate secrets nested arbitrarily deep in values. It also enforces synchronization on certain shared maps.


iwahbe:
To speed up the port of #38, I'm merging without external approval.

PR #47, opened 2022-08-23
Comply with invoke naming conventions

iwahbe:
It looks like invokes should start with a lower case letter. This PR brings inferred schemas into compliance.


aq17 (approved):
LGTM

Issue #39, opened 2022-08-15
Port a component provider: pulumi-kubernetes-coredns


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Port the relatively small component provider https://github.com/pulumi/pulumi-kubernetes-coredns.

iwahbe:
PR is https://github.com/pulumi/pulumi-kubernetes-coredns/pull/6.

PR #48, opened 2022-08-25
Improve handling of foreign types

iwahbe:
This PR makes the `type="..."` annotation more powerful. It also omits generating subtypes for types marked as foreign (but not actually foreign).

This is necessary for #39.


iwahbe:
For reviewers: most of the changes are from `go mod tidy`. The changes made are small.

aq17 (approved):
LGTM but may want @Zaid-Ajaj to take a look first if he's more familiar with this component

PR #50, opened 2022-08-29
Allow merging with the schema middleware

iwahbe:
Here we allow the schema provider to call `GetSchema` to underlying providers. A top level merge is performed, with the calling `schema.Provider` getting priority.


aq17 (approved):
👍🏼

PR #51, opened 2022-08-30
Usability improvements

iwahbe:
- Remove semver from the public interface.
- Add a dedicated top-level GetSchema function that operates on a provider.


aq17 (approved):
LGTM

PR #52, opened 2022-09-02
Move Provider type to a struct based approach

iwahbe:
I'm testing this in response to a comment from @lukehoban.

You can look at the examples to get a sense of the change.

If you think this is a good or bad change, add a comment explaining why.


iwahbe:
@justinvp This is ready for review.

aq17 (approved):
LGTM 👍🏼

examples/credentials/main.go
```diff
@@ -18,12 +18,13 @@ func main() {
 }
 
 func provider() p.Provider {
-	return infer.NewProvider().
-		WithResources(infer.Resource[*User, UserArgs, UserState]()).
-		WithModuleMap(map[tokens.ModuleName]tokens.ModuleName{
+	return infer.Wrap(p.Provider{}, infer.Options{
```
justinvp, line 21: It seems likely that there will be providers that are implemented entirely with `infer` and having to do `infer.Wrap(p.Provider{}, infer.Options{...})` isn't great usage.

Instead, could we have `infer.NewProvider(infer.Options{...})` or even `infer.Provider(infer.Options{...})`?

If we added that, would we still need `Wrap` to wrap an existing `p.Provider`?

justinvp (approved):
Let's use the new `infer.Provider` helper in the examples. Aside from that, LGTM. Thanks.

PR #55, opened 2022-09-12
Pass provider.Context to infer.Component.Construct

### LINKED ISSUE #54: Pass a `provider.Context` to `infer.Component.Constuct`
## Hello!

## Issue details

iwahbe:
Fixes #54 

I'm not aware of a way to unify the `*pulumi.Context` and `provider.Context` types such that a new type `*pulumi.Context&provider.Context` can be passed in place of either. For this, we would need `*pulumi.Context` to be an `interface` instead of a `struct`.

I'm open to other ideas here.

### DIFF
--- a/examples/random-login/main.go
+++ b/examples/random-login/main.go
@@ -48,7 +48,7 @@ type MoreRandomPasswordState struct {
 	Password *random.RandomPassword `pulumi:"password" provider:"type=random@v4.8.1:index/randomPassword:RandomPassword"`
 }
 
-func (r *MoreRandomPassword) Construct(ctx *pulumi.Context, name, typ string, args MoreRandomPasswordArgs, opts pulumi.ResourceOption) (*MoreRandomPasswordState, error) {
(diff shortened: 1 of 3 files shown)


justinvp:
I've been thinking about this. Now that I'm seeing it, it's a bit unfortunate to have to pass the two different contexts in, especially since I suspect the need for the `provider.Context` would be relatively rare.

Two alternative ideas to consider:

1. Augment `pulumi.Context` with general storage mechanism for extra data (key/values), and then leverage that to store away the `provider.Context`. That way, it'd hidden for the 90% cases where it's not needed, but would still be relatively easily retrievable in the 10% of cases where it's needed. Maybe accessed via `ctx.GetData("providerContext")` or something like that? Or maybe instead of the providerContext itself, we'd throw just the provider config, e.g. `ctx.GetData("providerConfig")`.2. 

2. Have a special `provider.ComponentContext` that has an extra `Context` field that returns the `pulumi.Context`. But then usage inside `Construct` would look like `ctx.Context.RegisterComponentResource(...)`. The usage is not ideal.

iwahbe:
Spike on embedding the the config data in the `pulumi.Context` context field. We also considered adding a `pulumi.Context.WithValue` that then calls `pulumi.Context.ctx = context.WithValue(pulumi.Context.ctx, V)`.

iwahbe:
Closing in favor of #56

Issue #53, opened 2022-09-12
Unify the Provider and Config section


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The inferred provider resource should populate the `config` as well as the `provider` section of the schema. Inputs and outputs should work as expected.

PR #56, opened 2022-09-14
Allow components to access config

### LINKED ISSUE #54: Pass a `provider.Context` to `infer.Component.Constuct`

iwahbe:
Fixes #54 

This system doesn't require passing 2 configs.


iwahbe:
Tests coming later.

infer/provider.go
```diff
@@ -62,16 +63,28 @@ func NewProvider() *Provider {
...
-	cv := ctx.Value(configKey)
+	v := ctx.Value(configKey)
+	return getConfig[T](v, v != nil)
+}
+
+// Retrieve the configuration of this provider.
+//
+// Note: GetComponentConfig will panic if the type of T does not match the type of the config or if
+// the provider has not supplied a config.
+func GetComponentConfig[T any](ctx *pulumi.Context) T {
```
justinvp, line 77: In addition to being able to access the provider config from component factories, I could imagine component authors wanting to access other provider-related information from the `p.Context` (e.g. being able to call the `Log` methods, etc.).

Should we provide a way to convert a `*pulumi.Context` from `Construct` into a `p.Context`? Effectively a function like `p.CtxFromPulumiContext(ctx *pulumi.Context) p.Context` that would live next to the other `Ctx` functions?

Then we could choose to not expose `GetComponentConfig` if we wanted. We could just have `GetConfig`, and component authors would have to do something like:

```go
// Get provider config.
pctx := p.CtxFromPulumiContext(ctx)
config := infer.GetConfig[Config](pctx)

// And do other things with the provider context.
pctx.Log(...)
```

If we did still want a helper like `GetComponentConfig` we could include it and it could be implemented like:

```go
func GetComponentConfig[T any](ctx *pulumi.Context) T {
	pctx := p.CtxFromPulumiContext(ctx)
	return GetConfig[T](pctx)
}
```

iwahbe: I think this is good future-proofing. 

You have access to `pulumi.Context.Log` already, so I don't think this is necessary for logging.

infer/component.go
```diff
@@ -98,3 +100,16 @@ func (rc *derivedComponentController[R, I, O]) Construct(pctx p.Context, typ str
...
 	return res, err
 }
+
+type componentContextKey struct{}
+
+// Retrieve a provider.Context from a pulumi.Context.
+//
+// This function is only valid when the *pulumi.Context was passed in from the infer
+// library.
+func CtxFromPulumiContext(ctx *pulumi.Context) p.Context {
```
justinvp, line 110: Should we move this to be next to the other `CtxXXX` funcs in `provider.go`?

That should avoid the need to have `embeddedData`: we wouldn't have to publicly export `GetEmbeddedData` and `PutEmbeddedData`, and wouldn't have to have `putEmbeddedMap` sprinkled throughout.

See my comment below on the implementation of `Construct` in `provider.go` below.

provider.go
```diff
@@ -832,7 +839,7 @@ func (p *provider) Delete(ctx context.Context, req *rpc.DeleteRequest) (*emptypb
 }
 
 func (p *provider) Construct(pctx context.Context, req *rpc.ConstructRequest) (*rpc.ConstructResponse, error) {
-	return comProvider.Construct(pctx, req, p.host.EngineConn(),
+	return comProvider.Construct(putEmbeddedMap(pctx), req, p.host.EngineConn(),
```
justinvp, line 842: If we moved `CtxFromPulumiContext` into this file, I think we could remove `embeddedData` and associated publicly exported `GetEmbeddedData` and `PutEmbeddedData`.

Then, the implementation here would look something like the following.

There'd be a `type pkgContextKey struct{}` key.

And inside `Construct` here would look something like:

```go
func (p *provider) Construct(pctx context.Context, req *rpc.ConstructRequest) (*rpc.ConstructResponse, error) {
	packageContext := &pkgContext{}
	pctx = context.WithValue(pctx, pkgContextKey{}, packageContext)
	packageContext.Context = pctx

	return comProvider.Construct(pctx, req, p.host.EngineConn(),
	...
}
```

And the implementation of `CtxFromPulumiContext` would look something like:

```go
func CtxFromPulumiContext(ctx *pulumi.Context) p.Context {
	v := ctx.Context().Value(pkgContextKey{})
 	contract.Assertf(v != nil,
 		"CtxFromPulumiContext must be called on the pulumi.Context passed in to Construct")
 	return v.(p.Context)
}
```

examples/random-login/main_test.go
```diff
@@ -17,7 +17,18 @@ const schema = `{
   "name": "random-login",
   "version": "0.1.0",
   "config": {},
```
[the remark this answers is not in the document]
iwahbe: Separate issue: https://github.com/pulumi/pulumi-go-provider/issues/53, but yes.

provider.go
iwahbe: I removed `GetEmbeddedData(ctx, key)` function. You can now just call `ctx.Value(key)`. I'm concerned with pre embedding the `provider.Context`, because middleware layers might change it.

iwahbe:
I think that #59 is a better solution to the problem this PR solves.

iwahbe:
Closed in favor of #59.

PR #59, opened 2022-09-19
Redesign construct api

### LINKED ISSUE #54: Pass a `provider.Context` to `infer.Component.Constuct`

iwahbe:
This PR redesigns the `Provider.Constuct` API. The new API allows deferring the construction of a `*pulumi.Context` until it is used. This obviates the need for mutable `context.Context`s. It also brings the API more inline with our other methods, the method taking a `(p.Context, p.ConstructRequest)` and returning a `(p.ConstructResponse, error)`.

There is no change to the `infer.Component` API.

The PR also enhances how examples are tested. Instead of just compiling them and running unit tests, we also run `pulumi up; pulumi up --expect-no-changes; pulumi destroy`. 

Fixes #54


examples/str/consumer/Pulumi.yaml
```diff
@@ -1,30 +1,36 @@
...
+plugins:
+  providers:
+    - name: str
+      path: ..
+
 variables:
   replaced:
     Fn::Invoke:
-      Function: str:Replace
+      Function: str:replace
```
AaronFriel, line 12: Is this because the str provider or YAML supports both casings, or because this provider is using a lowercase token now and requires it?

Asking as this appears inconsistent with tokens used elsewhere which - for whatever reason - capitalize the type at the end.

middleware/cancel/cancel.go
```diff
@@ -150,12 +148,10 @@ func Wrap(provider p.Provider) p.Provider {
...
 	if provider.Construct != nil {
-		new.Construct = func(pctx p.Context, typ string, name string,
-			ctx *pulumi.Context, inputs pprovider.ConstructInputs, opts pulumi.ResourceOption,
-		) (pulumi.ComponentResource, error) {
-			pctx, end := cancel(pctx, noTimeout)
+		new.Construct = func(ctx p.Context, req p.ConstructRequest) (p.ConstructResponse, error) {
+			ctx, end := cancel(ctx, noTimeout)
 			defer end()
-			return provider.Construct(pctx, typ, name, ctx, inputs, opts)
+			return provider.Construct(ctx, req)
```
AaronFriel, line 154: Good simplification of the interface.

examples/str/consumer/Pulumi.yaml
iwahbe: The `str` provider only supports a single casing, which is lowercase for functions. Pulumi YAML doesn't change anything there.

I changed functions to expose a lowercase name (`pkg:module:name`) since that is consistent with function type tokens in other packages. Function tokens follow a different convention than other resource tokens. I'm not sure why. This change was made a while ago, and the example was broken. Since it wasn't tested, I didn't notice. This PR adds tests to the examples.

PR #105, opened 2023-08-07
Make outputs computed when appropriate

### LINKED ISSUE #75: method to make field Computed?
## Hello!

## Issue details

I was working on https://github.com/pulumi/pulumi-command/issues/166. Afaik there's not a way to indicate that a field is computed. I believe that the provider is returning an empty string.

Some ideas:
- [ ] make properties on resources computed by default
- [ ] extend WireDependencies

iwahbe:
An output is marked as computed if:
1. Any of its inputs are computed.
2. Any of its inputs have changed. (assuming old inputs are available)
3. It is a preview before a create.

This property can always be overridden by setting `matcher.OutputField(&f.field).Known()`.

We don't allow users to always set an input as computed because it will cause constant diffs. Instead, we respect dependencies so they are set as diffed when appropriate.

Fixes #75


infer/resource.go
```diff
@@ -211,54 +214,127 @@ type fieldGenerator struct {
...
+		// unable to perform old-vs-new diffing here.
+		//
+		// This may lead to applies running on old information during
+		// preview. This is possible anyway, if user's dependencies don't
+		// accurately reflect their logic. This is not a problem for non-preview
+		// updates.
+		//
+		// The solution is to require embedding input structs in output structs
+		// (or do it for the user), ensuring that we have access to information
+		// that changed..
```
iwahbe, line 286: This PR does not exacerbate this problem. Diffing inputs is currently impossible unless they are embedded, since we don't preserve the old inputs.

infer/resource.go
```diff
@@ -289,7 +365,7 @@ func (g *fieldGenerator) InputField(a any) InputField {
 }
 
 func (g *fieldGenerator) OutputField(a any) OutputField {
-	if allFields, ok, err := g.argsMatcher.TargetStructFields(a); ok {
+	if allFields, ok, err := g.stateMatcher.TargetStructFields(a); ok {
```
iwahbe, line 368: This was a simple typo that testing missed. We didn't check for errors on the default case, and so we would generate an error on the default case always (and then ignore it). The assert on L835 prevents this from happening again.

t0yv0:
I'll need some overview of pulumi-go-provider to be able to review this. What's the context here.

resource.Computed seems like an old-school representation of unknowns before first-class outputs were available, can we just use first-class outputs instead?

In terms of marking outputs as unknown:

```
Any of its inputs are computed.
Any of its inputs have changed. (assuming old inputs are available)
It is a preview before a create.
```

This is a lot of policy decisions here the framework is making! Would this be right for all providers using the framework? It sounds a little suspect. Ultimately only the provider itself knows best which outputs are unknown and which are known based on partially known inputs and ideally there's an option for the provider to compute exactly that; is there? Perhaps simpler providers opt out of seeing unknowns and let the framework perform heuristics like the above one to compute them?

iwahbe:
> I'll need some overview of pulumi-go-provider to be able to review this. What's the context here.

This applies to inferred (from code) custom resources, so they are unable to see if their inputs are are computed or not. They see raw `T`, not `pulumi.Output[T]` or `resource.PropertyValue`. Each resource may specify, for each output field, which input fields they depend on. This is used when computing changes.

> 
> resource.Computed seems like an old-school representation of unknowns before first-class outputs were available, can we just use first-class outputs instead?

I wasn't aware of this, and `resource.Computed` isn't deprecated. I can convert to `resource.Output`. Unless you can find where the engine guarantees not to send `resource.Computed` values, I think I will need to handle both.

> 
> In terms of marking outputs as unknown:
> 
> ```
> Any of its inputs are computed.
> Any of its inputs have changed. (assuming old inputs are available)
> It is a preview before a create.
> ```
> 
> This is a lot of policy decisions here the framework is making! Would this be right for all providers using the framework? It sounds a little suspect. Ultimately only the provider itself knows best which outputs are unknown and which are known based on partially known inputs and ideally there's an option for the provider to compute exactly that; is there? Perhaps simpler providers opt out of seeing unknowns and let the framework perform heuristics like the above one to compute them?

For each set of checked inputs, they specify dependencies between inputs and outputs. They can also mark an output as `AlwaysKnown()`, overriding this action. As I mentioned in the PR body, we don't allow users to set `AlwaysComputed()` because it leads to perpetual diffs.

Providers that are not built on infer see raw `resource.PropertyValue` and can flow computedness through however they want.

t0yv0:
Oh okay that makes some sense! Remind me on Monday I'll dig in further into this and will be able to approve; unless you find another reviewer first.

AMKamel:
@t0yv0 Monday reminder

Makefile
```diff
@@ -9,6 +9,7 @@ test: test_unit test_examples
 .PHONY: test_unit
 test_unit: build
 	go test ./...
+	cd infer/tests && go test ./...
```
t0yv0, line 12: Nice catch.

infer/resource.go
```diff
@@ -211,54 +214,127 @@ type fieldGenerator struct {
...
+		// preview. This is possible anyway, if user's dependencies don't
+		// accurately reflect their logic. This is not a problem for non-preview
+		// updates.
+		//
+		// The solution is to require embedding input structs in output structs
+		// (or do it for the user), ensuring that we have access to information
+		// that changed..
+		oldInput, hasOldInput := oldInputs[k]
+		if inputs[k].IsComputed() || (hasOldInput && !inputs[k].DeepEquals(oldInput)) {
+			return resource.MakeComputed(prop)
```
t0yv0, line 289: I'm not sure I understand why this is necessary. If an input `k` is unknown, how will it ever be DeepEquals to an oldInput? Presumably oldInput is coming from state so it's never unknown.

infer/resource.go
```diff
@@ -211,54 +214,127 @@ type fieldGenerator struct {
...
-	// Create mandatory secrets
-	for s := range g.alwaysSecret {
-		s := resource.PropertyKey(s)
-		if m[s].IsComputed() {
-			break
-		}
-		if !m[s].IsSecret() {
-			v := m[s]
-			m[s] = resource.NewSecretProperty(&resource.Secret{Element: v})
+	if input, ok := inputs[key]; ok && !input.IsComputed() && input.DeepEquals(prop) {
```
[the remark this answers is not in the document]
t0yv0: OK we walked through this a little bit. The case @iwahbe made for marking things unknown when declared as such because the system is essentially slightly cheating /approximating propagation of unknowns. When passing data into the user code, unknowns are represented as zero values and the code is called anyway. When the output code received from the user code has zero values in unknown position it's' reasonable to assume these are unknowns. When the output code has non-zero values, they might have been a result of a computation like `x+y` that should have been `unknown+y=unknown` but actually works as `empty+y=y`, so that's the case for the framework discarding `y` and making things unknown when the static approximation specifies it should be unknown. 

The alternative we considered here is propagating this `y` out as known in spite of static specification to preserve more information.

Stepping back, this is heuristic land with no clear answers. What we'd like to have long term instead is an API for the user to hydrate PropertyValue to Go structures that's lossless with respect to unknown/empty distinction; basically if the provider author wants to handle unknowns then they bind to Go structures that can represent unknowns; and then if an incoming PropertyValue fails to bind to the Go structure then the framework either bypasses calling the user code all-together or errors out. We'll consider coming back to redesigning this pathway once there's more clarity on Go generics support in Pulumi..

t0yv0 (approved):

Issue #106, opened 2023-08-23
Spurious diff on secret inputs on the associated output field


iwahbe:
### What happened?

I noted during testing of https://github.com/pulumi/pulumi-command/pull/251 that we have spurious diffs on outputs that correspond to secret inputs. This is because with https://github.com/pulumi/pulumi-go-provider/pull/105, we have fixed the default WireDependencies implementation.

This introduced the following behavior with default WireDependencies:

Consider these inputs:

```
input: {
  public: "foo"
  private: secret("bar")
}

output: {
  combined: public+private
  public: public,
  private: private,
}
```

Default behavior is all outputs are dependent on all inputs. That means that `output.public` is made secret, since it depends on `private`, which is secret.

However, diff compares the old outputs with new inputs. That means that we compare `output.public` (which is secret) with `newInput.public` (unchanged => not secret). Since only one value is secret, we detect a change.

### Expected Behavior

`pulumi up --yes && pulumi preview --expect-no-changes` works in the default case.

### Steps to reproduce

Build pulumi-command against v0.11.0 of the SDK and delete the explicit `WireDependencies`.
Then run `pulumi up && pulumi preview` on this program:

```ts
const cmd = new local.Command("json", {
  create: `echo '{"json": true}'`,
  environment:pulumi.secret({
    "FOO": "bar",
  }),
});
```

### Output of `pulumi about`

NA

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #107, opened 2023-08-23
Do not flow secretness on input values

iwahbe:
Fixes https://github.com/pulumi/pulumi-go-provider/issues/106 (see issue for mechanics of the fix)

### DIFF
--- a/infer/tests/create_test.go
+++ b/infer/tests/create_test.go
@@ -88,6 +88,41 @@ func TestCreate(t *testing.T) {
 		}, resp.Properties)
 	})
 
+	t.Run("unwired-secrets", func(t *testing.T) {
+		prov := provider()
+		sec := resource.MakeSecret
+		str := resource.NewStringProperty
+		resp, err := prov.Create(p.CreateRequest{
+			Urn: urn("Echo", "create"),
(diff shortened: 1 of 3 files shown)


infer/tests/create_test.go
```diff
@@ -88,6 +88,41 @@ func TestCreate(t *testing.T) {
 		}, resp.Properties)
 	})
 
+	t.Run("unwired-secrets", func(t *testing.T) {
```
mjeffryes, line 91: Should we test the condition where the input has the same name, but a different value in the output still creates a secret?

mjeffryes (commented):
This looks reasonable to me, but I don't know if I have enough context to fully understand the impact of this change.

infer/tests/create_test.go
iwahbe: We probably should. I'll add that test.

iwahbe: Added in https://github.com/pulumi/pulumi-go-provider/pull/107/commits/4907e778f086ddaabc7d310e88969e52dd59b02d.

mjeffryes (approved):

PR #114, opened 2023-08-29
Apply defaults in Check

iwahbe:
This is a quick fix for https://github.com/pulumi/pulumi-command/issues/248. In theory, providers will never see values without defaults applied. This contract is not well enforced, with variance between different language SDKs and how aggressively defaults permute through nesting.

This change applies defaults as part of the Check operation for `infer`ed providers. We get two main benefits here:
1. We insulate from buggy SDKs. Implementing default values for arbitrarily nested types has been inconsistent in the past. (fix) AFAIK, no SDK implements default values through arbitrary maps and arrays.
2. Allow in-memory tests to benefit from default values by passing through values `Check`.

(1) solves a P1 now. (2) dramatically simplifies testing providers with default values.


infer/resource.go
```diff
@@ -547,11 +547,17 @@ func (rc *derivedResourceController[R, I, O]) Check(ctx p.Context, req p.CheckRe
...
 	// that we can de-serialize correctly
 	var i I
-	_, err := decode(req.News, &i, req.News.ContainsUnknowns())
+	secrets, err := decode(req.News, &i, req.News.ContainsUnknowns())
 	if err == nil {
+		if err := applyDefaults(&i); err != nil {
+			return p.CheckResponse{}, fmt.Errorf("unable to apply defaults: %w", err)
+		}
+
+		inputs, err := encode(i, secrets, false)
```
t0yv0, line 556: This only edits Check so I'm guessing defaults for Invokes and for the provider itself (CheckConfig) are not covered by the PR?

infer/tests/provider.go
```diff
@@ -173,13 +173,134 @@ func (*WiredPlus) WireDependencies(f infer.FieldSelector, a *WiredInputs, s *Wir
...
+type NestedDefaults struct {
+	// Direct vars. These don't allow setting zero values.
+	String string  `pulumi:"s,optional"`
+	Float  float64 `pulumi:"f,optional"`
+	Int    int     `pulumi:"i,optional"`
+	Bool   bool    `pulumi:"b,optional"`
+
+	// Indirect vars. These should allow setting zero values.
+	StringPtr *string  `pulumi:"ps,optional"`
+	FloatPtr  *float64 `pulumi:"pf,optional"`
```
t0yv0, line 223: Curious, are these types permitted here never Output or Input types? The reflective pass does not deal with Output/Input which makes it simple, but I was curious if that's warranted?

infer/apply_defaults.go
```diff
@@ -0,0 +1,354 @@
...
+					return false, nil
+				}
+				copy = copy.Elem()
+			}
+		}
+		didSet, err := d.apply(derefNonNil(s))
+		if err != nil {
+			return false, err
+		}
+		if didSet {
```
t0yv0, line 238: didSet tracking is optimizing away something here?

infer/tests/provider.go
iwahbe: They are never input or output types. The structs for custom resources deal exclusively in plain values.

t0yv0 (approved):
I think it's important for the provider to ensure defaults are applied so this LGTM! I had some questions but they're not blocking. Reviewing the code it looks reasonable I get a little lost proving the correctness of pointer case tracking in my head but going to trust the test cases here. I'm comfortable maintaining this code or fixing bugs if those arise here.

infer/resource.go
iwahbe: I was planning to follow up for those, but it would be better to make that explicit.
https://github.com/pulumi/pulumi-go-provider/issues/115
https://github.com/pulumi/pulumi-go-provider/issues/116

infer/apply_defaults.go
```diff
@@ -0,0 +1,354 @@
...
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	"github.com/pulumi/pulumi-go-provider/internal/introspect"
+)
+
+// The object that controls default application.
+type defaultsWalker struct {
+	// seen is the stack of types that defaultsWalker has descended into.
+	seen []reflect.Type
```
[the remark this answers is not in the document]
iwahbe: This is expressed on the `reflect.*` domain because annotations are defined on go types, not on property values. As an aside, this makes it relatively to handle recursive types, since go types already define the properties we want to bound recursion.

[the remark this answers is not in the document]
iwahbe: If we switched to applying defaults to property values, we would still need the recursive traversal to gather the default values, but would then need to draw another mechanism in to apply the values.

infer/apply_defaults.go
iwahbe: We explore the full space of our input's fields to find where defaults should be applied, but we don't want to set up structs that were `(*struct{...})(nil)` unless the struct has a default value that should be applied. The pseudocode looks like this:

```go
if typeOf(field) == *T && field == nil {
    var copy &T{}
    didSet := apply(copy)
    if didSet {
        field = copy
    }
    // Otherwise we leave field to be nil
}
```

infer/apply_defaults.go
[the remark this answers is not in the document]
iwahbe: This behavior is under test because of this field:

https://github.com/pulumi/pulumi-go-provider/blob/9da507871298b1d0cf80626437761a0f3b32286a/infer/tests/provider.go#L201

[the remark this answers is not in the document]
t0yv0: >  would then need to draw another mechanism in to apply the values.

It would seem that if you're generating default values into Pulumi schema there's already a mechanism somewhere or that, but ymmv, I'm not familiar with the codebase so taking your word for it.

PR #117, opened 2023-08-31
Ignore unknown fields in Configure and CheckConfig

### LINKED ISSUE #110: Errors with custom provider config
### What happened?

I am having troubles implementing a custom provider config. When using the provider (eg. `pulumi up`), I get error messages like these ones, indicating that `version` and `pluginDownloadURL` should be part of `Config`:

```
    error: pulumi:providers:aci resource 'default_0_0_0_github_/api.github.com/netascode': property version value {0.0.0} has a problem: Unrecognized field 'version' on 'aci.Config'
    error: pulumi:providers:aci resource 'default_0_0_0_github_/api.github.com/netascode': property pluginDownloadURL value {github://api.github.com/netascode} has a problem: Unrecognized field 'pluginDownloadURL' on 'aci.Config'
```

As a workaround I have added the attributes to Config:

```
type Config struct {
	Version           *string `pulumi:"version,optional"`
	PluginDownloadURL *string `pulumi:"pluginDownloadURL,optional"`
}
```

Though I guess this should not be required.

The full code can be found here: https://github.com/netascode/pulumi-aci/blob/main/provider/pkg/provider/apic/config.go

### Expected Behavior

Custom config to work without adding `version` and `pluginDownloadURL` to `Config` struct.

### Steps to reproduce

Code and steps linked in description above.

### Output of `pulumi about`

CLI          
Version      3.78.1
Go Version   go1.20.7
Go Compiler  gc

Host     
OS       darwin
Version  13.4.1
Arch     arm64

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #110


t0yv0:
I've scrutinized this it does what it says on the tin.

t0yv0 (approved):
I don't want to block the PR on this but I have a question on what's the plan for this codebase to deal with the oddity of JSON-in-proto encoding present in the protocol?

- does is surface to the users of pulumi-go-provider currently or they're shielded?
- in particular do nested structs work in Configure/CheckConfig? what about unknown/secrets in the nested case?

If we're shielding the user from this can we shield the tests also?

Perhaps I'm being tripped up by this:


```
type ConfigCustom struct {
	Number string `pulumi:"number,optional"`
	Parsed float64
}
```

Why isn't number a float64 already in the model, can it be float64 and the plumbing correctly decode JSON-in-proto resource.PropertyValue("42") to the number? Does decoding need to happen in the user-provided layer?

iwahbe:
> I don't want to block the PR on this but I have a question on what's the plan for this codebase to deal with the oddity of JSON-in-proto encoding present in the protocol?
> 
> * does is surface to the users of pulumi-go-provider currently or they're shielded?
> * in particular do nested structs work in Configure/CheckConfig? what about unknown/secrets in the nested case?

We shield `infer` users from this. I don't have tests for nested values, but it works in theory. I'll add an issue to test.

> 
> If we're shielding the user from this can we shield the tests also?
> 
> Perhaps I'm being tripped up by this:
> 
> ```
> type ConfigCustom struct {
> 	Number string `pulumi:"number,optional"`
> 	Parsed float64
> }
> ```
> 
> Why isn't Number a float64 already in the model, can it be float64 and the plumbing correctly decode JSON-in-proto resource.PropertyValue("42") to the number? Does decoding need to happen in the user-provided layer?

It seems like the test was unclear on what was under test. That's my bad. `Number` wasn't a `float64` so that `Configure` could have something to do. Decoding happens just like with normal resources. I have updated the test: `ConfigCustom` is now defined like this:

https://github.com/pulumi/pulumi-go-provider/blob/9ed88e1db02d9ca27779abcd8f152841386f446d/infer/tests/provider.go#L313-L316

PR #119, opened 2023-09-01
Replace default values application in config

### LINKED ISSUE #111: Adding a default value for a non-string config attribute fails
### What happened?

I would like to add a default value for a non-string config attribute using `Annotate()`:

```
type Config struct {
	Insecure          bool    `pulumi:"insecure"`
}

func (c *Config) Annotate(a infer.Annotator) {
	a.SetDefault(&c.Insecure, true, "ACI_INSECURE")
}
```

When using the provider (eg. `pulumi up`) I get the following error message:

```
    error: internal error: invalid property type: "boolean"
```

Removing the `a.SetDefault()` statement fixes the issue.

The full code is available here: https://github.com/netascode/pulumi-aci/blob/main/provider/pkg/provider/apic/config.go

### Expected Behavior

No error message.

### Steps to reproduce

Provided in description above.

### Output of `pulumi about`

CLI
Version 3.78.1
Go Version go1.20.7
Go Compiler gc

Host
OS darwin
Version 13.4.1
Arch arm64

### Additional context

_No response_

### Contributing

### LINKED ISSUE #116: Apply defaults to CheckConfigure
## Hello!

## Issue details

Following up on https://github.com/pulumi/pulumi-go-provider/pull/114, we should apply the same defaults strategy to CheckConfigure.

iwahbe:
This PR moves config level defaults application from `Configure` to `CheckConfig` and switches from an ad-hock implementation to the same `applyDefaults` function used for resource defaults.

Because of some uncertainty on if we correctly handle nested config values, I have tested this feature at the gRPC level, ensuring we are compatible.

Fixes #111 
Fixes #116


tests/grpc/config_test.go
```diff
@@ -0,0 +1,525 @@
...
+	replay "github.com/pulumi/pulumi-terraform-bridge/testing/x"
+	"github.com/stretchr/testify/require"
+
+	p "github.com/pulumi/pulumi-go-provider"
+	"github.com/pulumi/pulumi-go-provider/infer"
+	config "github.com/pulumi/pulumi-go-provider/tests/grpc/config/provider"
+)
+
+// These inputs were created by running `pulumi up` with PULUMI_DEBUG_GRPC=logs.json in
+// ./config/consumer.
```
mjeffryes, line 30: just to check if I'm following correctly: You created the consumer just for the purpose of capturing these test cases and captured these logs before making the change to infer/configuration.go to verify that the change did not affect what was sent over the wire right?

tests/grpc/config/schema-config.json
```diff
@@ -0,0 +1 @@
...ject","required":["s","b","i","m","a"]}},"provider":{"properties":{"a":{"type":"array","items":{"type":"string"}},"b":{"type":"boolean","defaultInfo":{"environment":["BOOL"]}},"db":{"type":"boolean","default":true},"di":{"type":"integer","default":42},"ds":{"type":"string","default":"defString"},"i":{"type":"integer","defaultInfo":{"environment":["INT"]}},"m":{"type":"object","additionalProperties":{"type":"string"}},"n":{"$ref":"#/types/config:index:ConfigNested"},"s":{"type":"string","defaultInfo":{"environment":["STRING"]}}},"required":["s","b","i","m","a","n"],"inputProperties":{"a":{"type":"array","items":{"type":"string"}},"b":{"type":"boolean","defaultInfo":{"environment":["BOOL"]}},"db":{"type":"boolean","default":true},"di":{"type":"integer","default":42},"ds":{"type":"string","default":"defString"},"i":{"type":"integer","defaultInfo":{"environment":["INT"]}},"m":{"type":"object","additionalProperties":{"type":"string"}},"n":{"$ref":"#/types/config:index:ConfigNested"},"s":{"type":"string","defaultInfo":{"environment":["STRING"]}}},"requiredInputs":["s","b","i","m","a","n"]},"resources":{"config:index:Get":{"properties":{"config":{"type":"string"}},"required":["config"]}}}
```
mjeffryes, line 1: Where does this file get used?

tests/grpc/config_test.go
```diff
@@ -0,0 +1,525 @@
...
+type GetArgs struct{}
+type GetState struct {
+	Config string `pulumi:"config"`
+}
+
+func (*Get) Create(ctx p.Context, name string, input GetArgs, preview bool) (string, GetState, error) {
+	config := infer.GetConfig[Config](ctx)
+	bytes, err := json.Marshal(&config)
+	return name, GetState{Config: string(bytes)}, err
+}
```
mjeffryes, line 519: I don't follow why we need a copy of this here? shouldn't it just be internal to config.Provider?

mjeffryes (approved):
Nifty test architecture; if it works the way I think it does, it seems like a pretty solid verification that the change works as expected.

tests/grpc/config/schema-config.json
iwahbe: It doesn't.

tests/grpc/config_test.go
iwahbe: Removed.

tests/grpc/config_test.go
iwahbe: I created the consumer just for the purpose of capturing these test logs. I did one capture before and verified that existing behavior worked as expected. Changing out to `applyDefaults` changes the behavior. Specifically it applies defaults in more cases and it does so correctly on more types.

PR #123, opened 2023-09-05
Separate computed and secret dependencies

### LINKED ISSUE #122: Wire secrets separate from dependency flow
## Hello!

## Issue details

`infer.ExplicitDependencies` provides a language for describing how inputs flow into outputs. We use this information in two ways:

1. We mark an output as computed during preview when an input has changed.
2. We mark an output as secret when an input is secret.

When dependencies are not explicit, we assume that the output depends on the input without any granularity.

Tying (1) and (2) together are a sensible default, but we need the ability to break the connection. For example:

> `command:remote:Command` has an input that is always secret: `connection`. The output of `command:remote:Command` always depends on `connection` (we should return computed during preview if `connection` changes). However, we don't want the secretes of `connection` to apply to the rest of `command:remote:Command`.

When this feature is implemented, users should be able to:
1. Get default behavior that works as expected:
   - (computed) all outputs depend on all inputs.
   - (secret) secrets only flow to types of the same name, unless otherwise specified.
2. Specify arbitrary outputs to depend on arbitrary inputs, providing different specifications for computedness and secretness.
3. Specify only computedness and get default secrets or vice versa.

iwahbe:
Fixes #122 

---

This PR separates dependency flow for computedness and secretness.

By default, we apply the following strategies:

### Secrets
An output property is secret if and only if there is an input property of the same name that is also secret

### Computed
An output property is computed if there is any change in between old and new inputs unless there is an input property of the same name and value that is not computed.

### DIFF
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -207,9 +207,16 @@ type OutputField interface {
 type InputField interface {
 	// Seal the interface.
 	isInputField()
+
+	// Only wire secretness, not computedness.
+	Secret() InputField
+
+	// Only flow computedness, not secretness.
+	Computed() InputField
 }
 
 type fieldGenerator struct {
+	args, state  any
 	argsMatcher  introspect.FieldMatcher
 	stateMatcher introspect.FieldMatcher
 	err          multierror.Error
@@ -232,10 +239,20 @@ type field struct {
 	// The set of tags that should never be secret
 	neverSecret bool
 	// A map from a field to its dependencies.
-	deps []string
+	deps []dependency
 
 	// If the output is known, regardless of other factors.
 	known bool
+}
+
+type dependency struct {
+	name string
+	kind inputKind
+}
+
+// Check if we should apply this dependency to a kind.
+func (d dependency) has(kind inputKind) bool {
+	return d.kind == inputAll || kind == d.kind
 }
 
 // MarkMap mutates m to comply with the result of the fieldGenerator, applying
@@ -271,7 +288,10 @@ func markComputed(
 
 	// If a dependency is computed or has changed, we mark this field as computed.
 	for _, k := range field.deps {
-		k := resource.PropertyKey(k)
+		if !k.has(inputComputed) {
+			continue
+		}
+		k := resource.PropertyKey(k.name)
 
 		// Not all resources embed their inputs as outputs. When they don't we are
 		// unable to perform old-vs-new diffing here.
@@ -322,7 +342,10 @@ func markSecret(
 	// Otherwise secretness is derived from dependencies: any dependency that is
 	// secret makes the field secret.
 	for _, k := range field.deps {
-		if ende.IsSecret(inputs[resource.PropertyKey(k)]) {
+		if !k.has(inputSecret) {
+			continue
+		}
+		if ende.IsSecret(inputs[resource.PropertyKey(k.name)]) {
 			return ende.MakeSecret(prop)
 		}
 	}
@@ -349,7 +372,7 @@ func (g *fieldGenerator) InputField(a any) InputField {
 			g.err.Errors = append(g.err.Errors, err)
 			return &errField{}
 		}
-		return &inputField{allFields}
+		return &inputField{fields: allFields}
 	}
 	field, ok, err := g.argsMatcher.GetField(a)
 	if err != nil {
@@ -396,8 +419,9 @@ func (g *fieldGenerator) OutputField(a any) OutputField {
 	return &errField{}
 }
 
-func newFieldGenerator[I, O any](i *I, o *O) *fieldGenerator {
+func newFieldGenerator(i, o any) *fieldGenerator {
 	return &fieldGenerator{
+		args: i, state: o,
 		argsMatcher:  introspect.NewFieldMatcher(i),
 		stateMatcher: introspect.NewFieldMatcher(o),
 		err: multierror.Error{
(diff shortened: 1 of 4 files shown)


infer/resource.go
```diff
@@ -207,9 +207,16 @@ type OutputField interface {
 type InputField interface {
 	// Seal the interface.
 	isInputField()
+
+	// Only wire secretness, not computedness.
```
t0yv0, line 211: This strikes me as very cryptic, can we unpack a bit?

infer/resource_test.go
```diff
@@ -16,15 +16,112 @@ package infer
...
+					ende.DeepEquals(newV, v) {
+					continue
+				}
+				assert.True(t, ende.IsComputed(v),
+					"key: %q", string(k))
+			}
+		} else if !ende.DeepEquals(
+			r.NewObjectProperty(oldInput),
+			r.NewObjectProperty(newInput)) {
+			// If there is a change, then every item item should be
```
t0yv0, line 102: Hmm, why is this? This is simulating preview only I guess but still?  It's a little interesting, trying to wrap my head around. So the implementation is allowed to copy oldInput to output, or mark outputs computed.

infer/resource_test.go
```diff
@@ -16,15 +16,112 @@ package infer
...
+			// computed, except items that mirror a known input.
+			for k, v := range output {
+				if n, ok := newInput[k]; !ok ||
+					ende.IsComputed(n) {
```
t0yv0, line 106: In the case of `!ende.IsComputed(n)` there's no restrictions placed at all on the output, is that desirable? The commend kind of made me think that the only behavior allowed is copying the known input.

infer/resource_test.go
```diff
@@ -16,15 +16,112 @@ package infer
...
+	type output struct {
+		input
+
+		O1 *string        `pulumi:"o1,optional"`
+		O2 float64        `pulumi:"o2"`
+		O3 map[string]int `pulumi:"o2"`
+	}
+
+	assert := func(t *testing.T, oldInput, newInput, output r.PropertyMap) {
+		if newInput.ContainsUnknowns() {
```
t0yv0, line 90: Doesn't newInput.ContainsUnknowns() make `ende.IsComputed(n)` impossible  on L106?

t0yv0:
I'll go read the ticket again https://github.com/pulumi/pulumi-go-provider/issues/122

t0yv0:
The ticket makes sense https://github.com/pulumi/pulumi-go-provider/issues/122 spec-wise actually I need some more time to work through the PR to grok it.

infer/resource_test.go
iwahbe: The default logic is: "if any input property has changed, declare all the output properties as unknown computed, except the outputs that are just pass-through inputs". We mark all output properties as unknown because we don't know if the changed input will change the output. We do know if passthrough inputs are known.

infer/resource_test.go
iwahbe: We don't require that an having an input and output share the same name implies they share the same value.

infer/resource_test.go
iwahbe: Yes, yes it does.

infer/resource_test.go
t0yv0: In this condition:

```go
if n, ok := newInput[k]; !ok || ende.IsComputed(n) {
```

If `newInput[k]` makes ok==false, then there are no asserts performed at all. 

So I don't think the property you state in tested here. But implementation might still do it right! It's just not tested seemingly, or I don't get something.

Consider oldInput={"a": 1}, newInput={"a": 42}, output: {"a": 42, "b": "foo"}.

This is trying to simulate "a" input changing, I think this will pass the check although "b": "foo" didn't get marked unknown. By your logic above it seems that "b" should be unknown since we can't assume that pushing changed inputs through the resource black box won't change the outputs.

t0yv0:
Don't want to block the Command P1 fix. I can't find anything wrong with it but I'm still a bit fluffy on the verifying the implementation, if not for the time crunch I'd like to go more thorough on clarifying the by-design behavior especially since secret / unknown inference through black boxes is something we do and want to do across many providers (definitely all bridged providers!) it's an area worth getting rock-solid some time.

t0yv0 (approved):

infer/resource_test.go
iwahbe: I have removed the `ende.IsComputed` check in https://github.com/pulumi/pulumi-go-provider/pull/123/commits/dd9e2fe939cbd7ea8b659239b060905dcb12af7c.

infer/resource_test.go
iwahbe: I think you're right that the test is missing something. https://github.com/pulumi/pulumi-go-provider/pull/123/commits/c8c23a07d70f783a5048af86d9c5d5b5fcd0ec24 strengthens the test to assert our expected conditions: 

https://github.com/pulumi/pulumi-go-provider/blob/c8c23a07d70f783a5048af86d9c5d5b5fcd0ec24/infer/resource_test.go#L99-L113

iwahbe:
> Don't want to block the Command P1 fix. I can't find anything wrong with it but I'm still a bit fluffy on the verifying the implementation, if not for the time crunch I'd like to go more thorough on clarifying the by-design behavior especially since secret / unknown inference through black boxes is something we do and want to do across many providers (definitely all bridged providers!) it's an area worth getting rock-solid some time.

I would love to work with you on nailing the semantics here, and then building a library that works the way we want it to, both here and in the bridge.

PR #124, opened 2023-09-06
Use supported go version

iwahbe:
Go only supports the last two minor version, which means only 1.20 and 1.21 are currently supported. This moves us to 1.20, and adds testing for 1.21.


guineveresaenger (approved):
thank you for the renamings!

PR #126, opened 2023-09-14
Don't hydrate nil structs that are marked as optional

iwahbe:
Consider input with this type:

```go
type Note struct {
	Title string  `pulumi:"title"`
        Body  *Nested `pulumi:"body,optional"`
}

type Body struct {
	Color string `pulumi:"color,optional"`
	Text  string `pulumi:"text"`
}

func (b *Body) Annotate(a infer.Annotator) {
	a.SetDefault(&b.Color, "black")
}
```

A `Note` must have a title and it may have a body. If it does have a body, then the body must have text. `Body.Color` defaults to `"black"`.

Consider what `applyDefaults` looks like for these values (before this PR):

1.
```ts
applyDefaults({"title": "do thing", "body": {"text": "due by friday"}})
=
{"title": "do thing", "body": {"text": "due by friday", "color": "black"}}
```

2.
```ts
applyDefaults({"title": "do thing", "body": {"text": "due by friday", "color": "blue"}})
=
{"title": "do thing", "body": {"text": "due by friday", "color": "blue"}}
```

3.
```ts
applyDefaults({"title": "do thing"})
=
{"title": "do thing", "body": {"color": "black"}}
```

---

(1) and (2) are correct. Importantly, both the input and output of (1) and (2) are valid values for `Note`. (3) shows the bug that this PR fixes: we start with a *valid* input, but after `applyDefaults` we have an *invalid* result, since `text` is required for `Body` but not present.

This PR prevents us from applying (and thus hydrating) nil pointers to optional structs. After this PR, we get a valid result on (3):

```ts
applyDefaults({"title": "do thing"}) = {"title": "do thing"}
```

### DIFF
--- a/infer/apply_defaults.go
+++ b/infer/apply_defaults.go
@@ -61,6 +61,7 @@ func (d *defaultsWalker) apply(v reflect.Value) (bool, error) {
 	// We get the set of default types that could be applied to v.
 	a := getAnnotated(t)
 	fields := map[string]reflect.Value{}
+	optional := map[string]bool{}
 	for _, field := range reflect.VisibleFields(v.Type()) {
 		tag, err := introspect.ParseTag(field)
 		if err != nil {
@@ -69,6 +70,8 @@ func (d *defaultsWalker) apply(v reflect.Value) (bool, error) {
 		if tag.Internal {
 			continue
 		}
+
+		optional[tag.Name] = tag.Optional
 		fields[tag.Name] = v.FieldByIndex(field.Index)
 	}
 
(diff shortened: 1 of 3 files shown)


thomas11:
Cases (1) and (3) seem inconsistent: in (1), the optional Body.Color is filled in, whereas in (3), the optional Note.Body is left out. Is that due to the `Annotate` method? Would (1) leave out Body.Color if `Annotate` was different or non-existent?

iwahbe:
> Cases (1) and (3) seem inconsistent: in (1), the optional Body.Color is filled in, whereas in (3), the optional Note.Body is left out. Is that due to the `Annotate` method? Would (1) leave out Body.Color if `Annotate` was different or non-existent?

Yes. `Annotate` is what says `Body.Color` has a default value: `"black"`. `Note.Body` does not have a default value.

thomas11 (approved):

PR #129, opened 2023-09-26
Allow override a Pulumi token by using SetToken of infer.Annotator

iwahbe:
This is a cleaned up version of https://github.com/pulumi/pulumi-go-provider/pull/128 (written by @tmeckel).

### DIFF
--- a/infer/function.go
+++ b/infer/function.go
@@ -57,9 +57,9 @@ func (*derivedInvokeController[F, I, O]) GetToken() (tokens.Type, error) {
 	//
 	//	pkg:index:FizzBuzz
 	//
-	// Functions use a different capitalization convection, so we need to convert:
+	// Functions use a different capitalization convention, so we need to convert:
 	//
-	//	pkg:index:FizzBuzz
+	//	pkg:index:fizzBuzz
 	//
 	return getToken[F](fnToken)
 }
--- a/infer/function.go
+++ b/infer/function.go
@@ -24,7 +24,6 @@ import (
 
 	p "github.com/pulumi/pulumi-go-provider"
 	"github.com/pulumi/pulumi-go-provider/infer/internal/ende"
-	"github.com/pulumi/pulumi-go-provider/internal/introspect"
 	t "github.com/pulumi/pulumi-go-provider/middleware"
 	"github.com/pulumi/pulumi-go-provider/middleware/schema"
 )
@@ -53,17 +52,16 @@ type derivedInvokeController[F Fn[I, O], I, O any] struct{}
 
 func (derivedInvokeController[F, I, O]) isInferredFunction() {}
 
-func (c *derivedInvokeController[F, I, O]) GetToken() (tokens.Type, error) {
-	var f F
-	annotator := getAnnotated(reflect.TypeOf(f))
-	if annotator.Token != "" {
-		return fnToken(tokens.Type(annotator.Token)), nil
-	}
-	tk, err := introspect.GetToken("pkg", f)
-	if err != nil {
-		return "", err
-	}
-	return fnToken(tk), nil
+func (*derivedInvokeController[F, I, O]) GetToken() (tokens.Type, error) {
+	// By default, we get resource style tokens:
+	//
+	//	pkg:index:FizzBuzz
+	//
+	// Functions use a different capitalization convection, so we need to convert:
+	//
+	//	pkg:index:FizzBuzz
+	//
+	return getToken[F](fnToken)
 }
 
 func fnToken(tk tokens.Type) tokens.Type {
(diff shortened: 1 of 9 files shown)


internal/introspect/annotator.go
```diff
@@ -83,3 +84,7 @@ func (a *Annotator) SetDefault(i any, defaultValue any, env ...string) {
 	a.Defaults[field.Name] = defaultValue
 	a.DefaultEnvs[field.Name] = append(a.DefaultEnvs[field.Name], env...)
 }
+
+func (a *Annotator) SetToken(module, token string) {
+	a.Token = fmt.Sprintf("pkg:%s:%s", module, token)
```
tmeckel, line 89: Should `SetToken` test the passed parameters in any way? For the `module` I'd check for traling/leading `:` and whitespaces. For the `module` I'd check at least for leading `:` and whitespaces. For the result of `fmt.Sprintf` I'd call `TrimSpace` to ensure for no trainglin whitespaces.

iwahbe: We can add that.

tmeckel: Like it. Makes the function more reliable. 👍🏼

infer/function.go
```diff
@@ -54,12 +53,15 @@ type derivedInvokeController[F Fn[I, O], I, O any] struct{}
...
-		return "", err
-	}
-	return fnToken(tk), nil
+	// By default, we get resource style tokens:
+	//
+	//	pkg:index:FizzBuzz
+	//
+	// Functions use a different capitalization convection, so we need to convert:
+	//
+	//	pkg:index:FizzBuzz
```
mjeffryes, line 62: The capitalization here is the same as above?

internal/introspect/annotator.go
```diff
@@ -83,3 +86,13 @@ func (a *Annotator) SetDefault(i any, defaultValue any, env ...string) {
 	a.Defaults[field.Name] = defaultValue
 	a.DefaultEnvs[field.Name] = append(a.DefaultEnvs[field.Name], env...)
 }
+
+func (a *Annotator) SetToken(module, token string) {
+	if !tokens.IsQName(module) {
+		panic(fmt.Sprintf("Module (%q) must comply with %s, but does not", module, tokens.QNameRegexp))
+	}
+	if !tokens.IsName(token) {
+		panic(fmt.Sprintf("Token (%q) must comply with %s, but does not", token, tokens.NameRegexp))
```
mjeffryes, line 95: If you get this wrong, when do you see the panic? At compile time for the provider?

infer/tests/token_test.go
```diff
@@ -0,0 +1,198 @@
...
+	panic("unimplemented")
+}
+
+type ObjectToken struct {
+	Value string `pulumi:"value"`
+}
+
+func (c *ObjectToken) Annotate(a infer.Annotator) { a.SetToken("obj", "Customized") }
+
+func TestTokens(t *testing.T) {
```
mjeffryes, line 77: It looks like this change touches the code that generates the default names; do we have a similar test that checks the names produced without these annotations?

internal/introspect/introspect.go
```diff
@@ -89,23 +87,14 @@ func GetToken(pkg tokens.Package, i any) (tokens.Type, error) {
 		typ = typ.Elem()
 	}
 
-	var name string
-	var mod string
-	if typ.Kind() == reflect.Func {
-		fn := runtime.FuncForPC(reflect.ValueOf(i).Pointer())
-		parts := strings.Split(fn.Name(), ".")
-		name = parts[len(parts)-1]
-		mod = strings.Join(parts[:len(parts)-1], "/")
```
mjeffryes, line 98: I didn't see this move anywhere else in this review; do we not want this behavior? is is covered somewhere else by the change in layering?

mjeffryes (commented):
Nifty!

internal/introspect/annotator.go
iwahbe: At schema generation time.

infer/tests/token_test.go
iwahbe: We have a bunch of integration tests that verify their schema.

infer/function.go
iwahbe: Fixed in https://github.com/pulumi/pulumi-go-provider/pull/129/commits/4d9db7c28a33a15140381ceb7a9363bb3c825040

internal/introspect/introspect.go
iwahbe: We never use this behavior. This was a remnant from when we tried to make Invoke and Construct calls wrappers around functions instead of structs. I forgot to remove it.

mjeffryes (approved):

PR #142, opened 2023-10-25
Apply defaults to invokes

### LINKED ISSUE #115: Apply provider side defaults to Invokes
## Hello!

## Issue details

Following https://github.com/pulumi/pulumi-go-provider/pull/114, we should apply defaults to invokes.

iwahbe:
Fixes #115 

This change applies default values as part of `Invoke`. This is inline with the behavior for resources.


mjeffryes (approved):
I wish I understood the general context better here, but the change seems reasonable and the tests pass.

iwahbe:
https://github.com/pulumi/pulumi-go-provider/pull/114 has the bulk of the discussion on why this is needed.

PR #158, opened 2023-12-04
Update to pulumi 3.95.0

MMartyn:
There was a change to the signature of the resource.NewURN method in this [change](https://github.com/pulumi/pulumi/commit/516979770f11d3a426239429731cf9f327c38836#diff-841290eccbd676c43597489825cc8f1628359283c634321e496e76e31f6fed9cR110) so needed to change some of the code. I wasn't sure if it should be token.QName(name).String() or just name. I opted for name as it was what I saw was done in the change to pulumi. Also the return type of Name() changed for URN as well [here](https://github.com/pulumi/pulumi/commit/516979770f11d3a426239429731cf9f327c38836#diff-21db7f7404c2baf3827464d874627369bde94bc36c8681f2b6b3908d4eeeab83L143) so had to update that in a few places.


iwahbe (approved):
Hey @MMartyn. Thanks for the PR!

PR #188, opened 2024-01-27
Enable attaching RawServer to existing HostClient

blampe:
I've been debugging panics in my logging and realized this is because `[RawServer](https://github.com/pulumi/pulumi-docker/pull/890/files#diff-cd386c4a0740a8677267cb5b3c15dd286eb87638949f649e5b0750803a2b7e48R30)` was initializing with a `nil` `HostClient`.

This essentially exposes `newProvider` so we can attach an existing `HostClient` to it. I only see this used in one test, so I opted for a breaking change instead of adding another function.

Refs https://github.com/pulumi/pulumi-docker/pull/890

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -383,16 +383,13 @@ func RunProvider(name, version string, provider Provider) error {
 	return pprovider.Main(name, newProvider(name, version, provider.WithDefaults()))
 }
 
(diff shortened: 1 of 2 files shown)


iwahbe (approved):
Sounds good.

PR #191, opened 2024-02-09
resourcex package

EronWright:
This PR implements a suite of utility functions for working with `resource.PropertyValue` and `resource.PropertyMap`.

See [resourcex/README.md](https://github.com/pulumi/pulumi-go-provider/blob/f739339097bc758e780608a3ff57775ca3395dac/resourcex/README.md) for more details.


resourcex/decode.go
```diff
@@ -0,0 +1,79 @@
...
+func decodeM(props resource.PropertyMap) map[string]interface{} {
+	obj := make(map[string]interface{})
+	for _, k := range props.StableKeys() {
+		key := string(k)
+		obj[key] = decodeV(props[k])
+	}
+	return obj
+}
+
+// decodeV returns a mapper-compatible object map, suitable for deserialization into structures.
```
EronWright, line 39: This function is similar to the venerable `MapRepl` function, which lacks recursion into secret and output values ([code](https://github.com/pulumi/pulumi/blob/ff6404a55f0d0e9c20ef94fcce2217e343145466/sdk/go/common/resource/properties.go#L611C10-L611C25)).

resourcex/extract.go
```diff
@@ -0,0 +1,122 @@
...
+type ExtractOptions struct {
+	// TagName is the struct tag name to use for resource property names; defaults to "json".
+	TagName string
+	// RejectUnknowns produces an error (of type ContainsUnknownsError) if any unknowns are extracted.
+	RejectUnknowns bool
+}
+
+type ExtractResult struct {
+	// Dependencies is the set of resources that the extracted information depends on (known or unknown).
+	Dependencies []resource.URN
```
iwahbe, line 33: I don't think we need to keep ordering here. I can't think of a situation where relying on the order of dependencies is correct.
```suggestion
	Dependencies map[resource.URN]struct{}
```

resourcex/traverse.go
```diff
@@ -0,0 +1,75 @@
...
+
+package resourcex
+
+import (
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+)
+
+// Traverse traverses a property value along the given property path,
+// invoking the given function for each property value it encounters.
+func Traverse(v resource.PropertyValue, p resource.PropertyPath, f func(resource.PropertyValue)) {
```
iwahbe, line 23: I can't think of an example where we want to call some function for every property described by a path. Can you provide some motivating examples for this design (as opposed to calling the function at the value described by the path).

resourcex/README.md
```diff
@@ -0,0 +1,243 @@
+# resourcex
+
+The `resourcex` package extends the `github.com/pulumi/pulumi/sdk/v3/go/common/resource` package with helpers
+for working with property values.
+
+1. `Extract` - Extract structured values from a property map, with tracking of unknownness and secretness.
+2. `Decode` - Decode a property map into a JSON-like structure containing only values.
```
iwahbe, line 7: I don't understand why we need both `Decode` and `Extract`. Shouldn't `Decode` be equivalent to:
```go
var dst map[string]any
_, err := Extract(props, &dst)
contract.AssertNoError(err)
return dst
```

resourcex/README.md
```diff
@@ -0,0 +1,243 @@
+# resourcex
+
+The `resourcex` package extends the `github.com/pulumi/pulumi/sdk/v3/go/common/resource` package with helpers
+for working with property values.
+
+1. `Extract` - Extract structured values from a property map, with tracking of unknownness and secretness.
```
iwahbe, line 6: We just had a design meeting where we agreed on a different way to implement `Extract` (where outputs/secrets are encapsulated in `pulumix.OutputOf[T]`). Is this yet another system for working unmarshalling?

Do we also plan to maintain this, or will we deprecate it?

resourcex/extract.go
```diff
@@ -0,0 +1,122 @@
...
+func Extract(target interface{}, props resource.PropertyMap, opts ExtractOptions) (ExtractResult, error) {
+	if opts.TagName == "" {
+		opts.TagName = "json"
+	}
+
+	// decode the property map into a JSON-like structure containing only values.
+	decoded := Decode(props)
+
+	// deserialize the JSON-like structure into a strongly typed struct.
+	config := &mapstructure.DecoderConfig{
```
iwahbe, line 51: Does this strategy mean that we will never be able to support targeting `pulumix.OutputOf[T]`, `resource.Asset` and `resource.Archive`?

resourcex/traverse.go
EronWright: The `Traverse` function is used by the `Extract` function to recover the unknownness and secretness of the extracted information. For example, an object value may itself be wrapped in a secret, contain fields that have values wrapped in secrets, etc.

resourcex/README.md
EronWright: I believe that we agreed in the meeting that `infer` is not ready to be adopted by the p/k provider, and that the provider should continue to use `resource.PropertyValue` as its primary abstraction. In keeping with that, this code is simply designed to provide views over `PropertyValue`.

resourcex/README.md
EronWright: `Extract` is implemented using `Decode` to obtain a map, then using the `mapstructure` package to map to a Go struct while summarizing the unknownness and secretness.

resourcex/extract.go
EronWright: The overarching idea is to create views of the values contained within a property map.  

Assets have a simple 'value' representation and may be decoded directly to `Asset`.  I implemented it in [a follow-up commit](https://github.com/pulumi/pulumi-go-provider/pull/191/commits/76bb7bdeaad8caff1cb763f503918e308f6d81a3).

Resource references are more interesting because `ResourceReference` has an `ID` field of type `PropertyValue`, which may be a string or be unknown.  In `Decode` one would emit an ID only if the value is known.  One could not extract a `ResourceReference` directly, but could extract a pure 'value' representation:
```
type ResourceReferenceValue struct {
  URN resource.URN `json:"urn,omitempty"`
  ID string `json:"id,omitempty"`
  PackageVersion string `json:"packageVersion,omitempty"`
}

Archives would require a similar 'value' representation.

File: resourcex/extract.go [hunk 7]
```diff
@@ -0,0 +1,122 @@
...
+	Dependencies []resource.URN
+	// ContainsUnknowns is true if the extracted information contains unknown values.
+	ContainsUnknowns bool
+	// ContainsSecrets is true if the extracted information contains secret values.
+	ContainsSecrets bool
+}
+
+// Extract extracts information from a property map into a target struct.
+// It returns a summary of the outputness and secretness of the extracted information.
+func Extract(target interface{}, props resource.PropertyMap, opts ExtractOptions) (ExtractResult, error) {
```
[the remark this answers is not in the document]
EronWright: I would like the function name to emphasize that it retrieves values and is lossy. Terms like "unmarshal" and "deserialize" might suggest a full-fidelity transformation. So I went with a term that is new to the codebase and would have a distinct meaning, i.e. "extract values from the property map".

resourcex/README.md
iwahbe: Let me rephrase, I don't understand why we need to make `Extract` and `Decode` public, since `Extract` provides a superset of `Decode`.

resourcex/extract.go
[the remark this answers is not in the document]
iwahbe: I disagree that `Unmarshal` implies a full-fidelity transformation. In Go, `Unmarshal` implies a lossy transformation from a untyped value to a typed representation. `json.Unmarshal` and `yaml.Unmarshal` are incredibly lossy. `json.Unmarshal` treats these scenarios as equivalent:

```go
var dst struct { S string `json:"s"` }
json.Unmarshal([]byte(`{}`), &dst)
json.Unmarshal([]byte(`{"s": ""}`), &dst)
json.Unmarshal([]byte(`{"f": "foo"}`, &dst)
```

resourcex/traverse.go
iwahbe: I might be confused. If that is the case and `.Metadata.Keys` is derived from `[1,2,3,4,5]`, doesn't that mean that `f` is called 5 times on the enclosing array?

---

`Extract` can call `traverse`, it doesn't necessitate making the function public.

t0yv0:
Thank you for this! The facilities address a big gap in provider development and I love seeing
progress here, and can approve to move things forward so we are not blocking on getting customer
value out. That said I believe that we should be able to do much better on the API, so if we merge
this it should be on the understanding that we make breaking changes and iterate on the API. Let me
limit the discussion to the interface here, as I believe it is premature to discuss implementation
until we are on the same page for the interface and sematics.

Using `go doc -all .` to get the interface:


```
FUNCTIONS

func Decode(props resource.PropertyMap) interface{}
    Decode decodes a property map into a JSON-like structure containing only
    values. Unknown values are decoded as nil, both in maps and arrays. Secrets
    are collapsed into their underlying values.

func Extract(target interface{}, props resource.PropertyMap, opts ExtractOptions) (ExtractResult, error)
    Extract extracts information from a property map into a target struct.
    It returns a summary of the outputness and secretness of the extracted
    information.

func IsContainsUnknownsError(err error) bool

func Traverse(v resource.PropertyValue, p resource.PropertyPath, f func(resource.PropertyValue))
    Traverse traverses a property value along the given property path, invoking
    the given function for each property value it encounters.


TYPES

type ContainsUnknownsError struct {
        Dependencies []resource.URN
}

func NewContainsUnknownsError(dependencies []resource.URN) *ContainsUnknownsError

func (e *ContainsUnknownsError) Error() string

type ExtractOptions struct {
        // TagName is the struct tag name to use for resource property names; defaults to "json".
        TagName string
        // RejectUnknowns produces an error (of type ContainsUnknownsError) if any unknowns are extracted.
        RejectUnknowns bool
}

type ExtractResult struct {
        // Dependencies is the set of resources that the extracted information depends on (known or unknown).
        Dependencies []resource.URN
        // ContainsUnknowns is true if the extracted information contains unknown values.
        ContainsUnknowns bool
        // ContainsSecrets is true if the extracted information contains secret values.
        ContainsSecrets bool
}
```

Decode looks suspect as "Unknown values are decoded as nil" is a suspect transformation, and Decode
applies it everywhere. Are there some motivating use cases that cannot be accomplished using other
primitives? Can this be a variation of Extract options, something definitely non-default (perhaps
named UnsafeDecodeUnknownAsNil)?

Extract should likely accept a PropertyValue not a PropertyMap to be more general. It can be fine to
provide a PropertyMap accepting variation for convenience.

Why does Extract accept target as a parameter? Would it be to awkward to return it in the result
instead, emphasizing the "direction" of the computation? I guess perhaps this is better because Go
can infer the desired type from the input parameter, and if we were returning a result we'd need to
accept another parameter that carries the desired type or perhaps a nil representative value. Maybe
this is the way to go.

Extract should have an inverse function to transform any to PropertyValue.

ContainsUnknownsError feels awkward (or at least redundant) and consider removing it as well as
removing RejectUnknowns option and IsContainsUnknownsError, NewContainsUnknownsError. This can be
accomplished by the user of using normal Extract and consulting ContainsUnknowns in the result, and
deciding to handle that. Perhaps it is included because it is a very frequent use case for the API?

ExtractResult is not specific to Extract but is an important type in the Pulumi Go SDK. Consider
this:

https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumi/internals/outputs.go#L41

```
type UnsafeAwaitOutputResult struct {
	Value        interface{}       // The value of the output. If unknown (in a dry-run), the value will be nil.
	Known        bool              // True if the value is known.
	Secret       bool              // True if the value is a secret.
	Dependencies []pulumi.Resource // The resources that this output depends on.
}

func UnsafeAwaitOutput(ctx context.Context, o pulumi.Output) (UnsafeAwaitOutputResult, error) {

```

And pulumix.Output[T] from:

https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumix/types.go#L53

I would love if these APIs could be reasonably unified and given helpful names. Sketching out,
something like this:

```
var pv resource.PropertyValue

var result pulumix.Output[MyStuff]

// func Extract[T any](PropertyValue, *pulumix.Output[T], ExtractOptions) error
err := Extract(pv, &result, opts)
// handle err

meta := UnsafeAwaitOutput(ctx, result)
meta.Known
meta.Secret
meta.Dependencies
```

This brings to the most important feature request. It would be lovely for Extract to recognize and
populate pulumix.Output[T] not only at the top level but also deeply inside the structures. This
would allow it to localize handling of secret/unknown metadata without loss of information.

This is in fact the most interesting and controversial bit here where some decisions need to be
made; what happens with information-losing transformations where information from PropertyValue does
not "fit" or have a representation in the desired output box of type T.

We have the following strategies:

  - abort the entire transformation with an error
  - go ahead and drop the information, possibly emitting a zero value of type T
  - float the information to the nearest enclosing pulumix.Output (which may happen to be top-level)

The float strategy seems very nice as it lets the user inject pulumix.Output at the right
granularity to contain the information loss. However one degenerate case here is pernicious,
floating an unknown all the way to top-level and losing all the information seems to be almost
always undesirable. Perhaps that should always be an error?

It would be lovely to give a discriminating user granular control of which strategy happens. But the
library also needs to decide which one is default, picking good defaults is hard.

I wonder if it would be practical to make the `drop` strategy opt-in while the default strategy is
float or abort. For example, this could have a form of permitting this with an annotation:

```
    type MyStruct struct {
       MyProperty string `pulumix:"mProperty,unknownAsEmpty"`
    }
```

Or it could take a form of a combinator on PropertyValue:

    pv = WithUnknownsAsNull(".myProperty", pv)

Finally I have some questions on tags, this looks a little peculiar:

```
type ExtractOptions struct {
        // TagName is the struct tag name to use for resource property names; defaults to "json".
        TagName string
```

I wonder why "json" is the default? Is this to facilitate reusing some structs that also double up
for JSON serialization? Seems like an unfortunate coupling? I would expect here a tag specific to
this library to be in use, no necessarily have that be configurable.

Now this API seems reasonable and entirely orthogonoal to the rest of things going on here:

```
func Traverse(v resource.PropertyValue, p resource.PropertyPath, f func(resource.PropertyValue))
```

I have a few more submissions of this same sort.

https://github.com/pulumi/pulumi-terraform-bridge/blob/master/unstable/propertyvalue/propertyvalue.go#L21

Transform generalizes traverse and I have a few more helpers like RemoveSecrets that are a special
form of Transform. I like these kinds of combinators and I like the clarity of multi-pass
transformations especially since we are working with small data where performance is not affected
much. But I have not yet secured alignment with the team to publish/reuse these. Perhaps we can
indeed incubate more of the PropertyValue helpers in pulumi-go-provider.

resourcex/extract.go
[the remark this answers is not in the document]
danielrbradley: I'd second this ... I think Unmarshal is what I'd look for in a Go library like this.

danielrbradley (approved):
I really like this. Would be good to get it in and get it field-tested. I expect this approach could sit alongside the infer based approach (which I expect will have fully Output support etc) as I think this will be easier to retrofit into existing providers but the infer approach might work well for larger reworking.

resourcex/extract.go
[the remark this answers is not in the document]
EronWright: I'll make the change.

resourcex/errors.go
```diff
@@ -0,0 +1,41 @@
...
+
+func NewContainsUnknownsError(dependencies []resource.URN) *ContainsUnknownsError {
+	return &ContainsUnknownsError{
+		Dependencies: dependencies,
+	}
+}
+
+func IsContainsUnknownsError(err error) bool {
+	return errors.Is(err, &ContainsUnknownsError{})
+}
```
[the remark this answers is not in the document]
EronWright: I expect that many usages of `Extract` would make use of the `RejectUnknowns` option (which produces ContainsUnknownsError). That's my rationale anyway for keeping this.

PR #195, opened 2024-03-01
id/urn are allowed on nested types

blampe:
https://github.com/pulumi/pulumi-go-provider/pull/160 had the unintended side effect of prohibiting "id" on _all_ types, but the engine only forbids those fields as outputs.

This PR changes our logic to only check top-level fields before drilling down. This still doesn't match the engine exactly, since technically these are allowed as inputs, but those edge cases seem unlikely.

### DIFF
--- a/infer/types.go
+++ b/infer/types.go
@@ -15,6 +15,7 @@
 package infer
 
 import (
+	"fmt"
 	"reflect"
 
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
Looks good to me.

PR #196, opened 2024-03-11
Add a public resourcex.DecodeValue function

mikhailshilkov:
I found `Decode` while re-implementing it in the AWS CC provider. (thank you @danielrbradley!)

I also need `DecodeValue` function there to be able to decode engine diffs. Therefore, adding it as a public member.


iwahbe (approved):
Looks good to me.

danielrbradley (approved):

Issue #205, opened 2024-03-27
[infer]: Prevent provider authors from describing optional fields as non-nullable types


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Not all zero values are valid for all languages. The inciting example is an enum in Python disallowing the `''` value: https://github.com/pulumi/pulumi-go-provider/issues/204. To work correctly, the provider needs to be able to distinguish a missing value and `T{}` on optional fields. 

The issue is most pressing for optional enums and optional structs. We should forbid fields like 👇 where `T` is of type `struct` or an enum.

```
fieldName T `pulumi:"name,optional"
```

blampe:
This would be extremely helpful!

It's not uncommon to use zero-valued structs in Go, but these generate "bad" schemas and there's no feedback loop so you're often left wondering "Should this be a pointer?"

iwahbe:
> This would be extremely helpful!
> 
> It's not uncommon to use zero-valued structs in Go, but these generate "bad" schemas and there's no feedback loop so you're often left wondering "Should this be a pointer?"

The problem isn't the generated schema, which is actually correct. The problem is that the Pulumie values are not round-trippable through the go structs. That's what caused the problem.

PR #206, opened 2024-03-27
Error during provider startup for invalid provider configurations

iwahbe:
Fixes https://github.com/pulumi/pulumi-go-provider/issues/205

This PR introduces checks for non-pointer (`T` instead of `*T`) values that cannot be safely round-tripped. So far, these include enum and struct values. Previously, users could specify a property like:

```
fieldName T `pulumi:"name,optional"`
```

This would cause confusing behavior at runtime if `T` transitively contained an enum without a pointer (or map, or array) indirection. To avoid a complicated rule set for where pointers are needed, we require a pointer where the `,optional` was declared.

### DIFF
--- a/infer/types.go
+++ b/infer/types.go
@@ -15,6 +15,7 @@
 package infer
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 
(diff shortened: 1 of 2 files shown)


t0yv0 (approved):
Looks good at first glance, the feature is very compelling, probably slightly breaking? If users used optional X before now this will force upgrading to *X. But I think it's worth it. Make illegal states unrepresentable.

I admit I'm not very thoroughly reviewing the code itself, LMK if it would be helpful to take the time to dig in. New codebase for me.

Issue #159, opened 2023-12-14
Consider removing provider.Context in favor of context.Context


blampe:
## Problem

https://github.com/golang/go/wiki/CodeReviewComments#contexts

> Don't create custom Context types or use interfaces other than Context in function signatures.
> 
> If you have application data to pass around, put it in a parameter, in the receiver, in globals, or, if it truly belongs there, in a Context value.

I wish this guidance went into more detail around _why_ to avoid custom Context types, but at least in our case the custom `provider.Context` type makes it difficult (impossible?) to unit test CRUD operations. 

Specifically, if a CRUD operation needs access to the provider's configuration it must use `infer.GetConfig` to get it. `GetConfig` requires a `provider.Context` from a `Wrap`'d provider, but once you have that provider you can't get a `provider.Context` from it because `ctx` is private. Instead, you need a server available (`ResourceProviderServer` or `integration.Server`) to generate a usable context.

https://github.com/pulumi/pulumi-go-provider/blob/ab334266917c79f43c0294e8a1602dc645c355ed/infer/provider.go#L129-L134

https://github.com/pulumi/pulumi-go-provider/blob/ab334266917c79f43c0294e8a1602dc645c355ed/infer/provider.go#L118-L120

https://github.com/pulumi/pulumi-go-provider/blob/ab334266917c79f43c0294e8a1602dc645c355ed/provider.go#L587-L589

In other words, as a user I want to be be able to unit test this function:

```go
func (*Resource) Delete(
    ctx provider.Context, 
    id string,
    state ResourceState, 
) error {
    cfg := infer.GetConfig[Config](ctx)
    ...
}
```

but I require a server to produce a valid `provider.Context`, so I'm actually only able to test at a higher layer:
```go
  err = server.Delete(provider.DeleteRequest{
      ID: "name",
      Urn: ...,
      Properties: resource.PropertyMap{...},
  })
  assert.NoError(t, err)
```

This is unfortunate because we lose one of the primary benefits of this library, namely the ability to work with simple structs instead of the lower-level `resource.PropertyMap`.

## Solution

One solution would be to export `provider.ctx(...)` → `provider.Ctx(...)` so users can create a `provider.Context` more easily. This is an _easy_ fix, but it still requires me to have a provider available, with configuration inferred, and so on. In other words, the framework is still "in the way," even for the simplest of test cases.

[...]

PR #207, opened 2024-03-28
Fix cancelation panic

### LINKED ISSUE #186: Cancellation panic
I was trying to test some cancellation behavior, so I ctrl-C'd during an update and got this:

```
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10581ba68]
    goroutine 27 [running]:
    github.com/pulumi/pulumi-go-provider/middleware/cancel.(*inOutCache[...]).drain(0x1070bb940?)
    	/Users/bryce/go/pkg/mod/github.com/pulumi/pulumi-go-provider@v0.14.0/middleware/cancel/cancel.go:229 +0x158
    github.com/pulumi/pulumi-go-provider/middleware/cancel.Wrap.func2({0x1070c2460, 0x14000d88720})
    	/Users/bryce/go/pkg/mod/github.com/pulumi/pulumi-go-provider@v0.14.0/middleware/cancel/cancel.go:56 +0x44
    github.com/pulumi/pulumi-go-provider.(*provider).Cancel(0x140008cc360, {0x1070b0778?, 0x14000d886f0}, 0x106e8a4e0?)
    	/Users/bryce/go/pkg/mod/github.com/pulumi/pulumi-go-provider@v0.14.0/provider.go:972 +0x8c
    github.com/pulumi/pulumi-docker/provider/v4.dockerHybridProvider.Cancel(...)
    	/Users/bryce/src/pulumi-docker/provider/hybrid.go:64
```

I'm not even sure if this a problem with go-provider, or if it's coming from a function closure around my own code.

AFAICT `provider.Main` is already performing a [GracefulStop](https://github.com/pulumi/pulumi/blob/f3e6e9a9dd9e12ae584bdc2f8f917ad4f31214f9/sdk/go/common/util/rpcutil/serve.go#L121) for us, so I don't understand what `inOutCache` is doing. Can we not simply pass through to the provider's `Cancel` implementation, if it has one?

iwahbe:
Fixes #186 

This PR is best reviewed by commit:

- 0ecbbcd98ecc4f24a3c6dbc04099f4a3aa545f4b is a pure refactor to improve the readability of the `cancel.Wrap` method. 
- 997d21b48a4f929889594e1c0b1c992eaacb4184 Replaces `inOutCache` with `evict.Pool`: A data structure better suited for purpose. `cancel` uses `evict.Pool` to keep track of the set of current requests, ensuring that the request associated with each context is canceled if one of three events happen:
  - The request returns.
  - The request times out (only for `Create`, `Update` and `Delete`).
  - The provider receives a `Cancel` request.
- 6d6b8aabff9bc4a3b8febc287635b88151d0b41b improves the integration testing around the `cancel` middleware, especially under heavy concurrency. I'm pleased to report that 6d6b8aabff9bc4a3b8febc287635b88151d0b41b without 997d21b48a4f929889594e1c0b1c992eaacb4184 regularly panics, showing that the improved testing does cover #186.
- 9a69c1fefb85581d7715a43f9519a88a219215b1 sets the `-race` flag for tests. This would have also caught #186.


blampe:
I'm still having trouble understanding why we need concurrency management here. Each request is handled in its own goroutine with its own context, so what is there to coordinate? I understand why we need some middleware to translate `context.Context` into `provider.Context` (https://github.com/pulumi/pulumi-go-provider/issues/159), but it seems like the `ctx.WithDeadline()` (or `p.CtxWithTimeout`) that we pass down should be sufficient?

> The provider receives a `Cancel` request.

It seems like this is primarily why we're trying to coordinate things across requests? If I understand, we get this `Cancel` and then try to _also_ cancel in-progress requests. Setting aside whether this is the right behavior (I don't think it is), calling [Stop](https://pkg.go.dev/google.golang.org/grpc#Server.Stop) on the server (which already has a global view of in-flight requests) _should_ accomplish the same goal. That is, each request would see a message on `ctx.Done()`. If that's _not_ the behavior we see, then that points towards a bug somewhere in how we propagate context.

Re: behavior, I don't think the `Cancel` handler's default behavior should impact other requests. If it _should_, I think that's a decision (and implementation) best left to the provider author. I would expect something more like: the first ctrl-C invokes `Cancel`, the server starts a graceful stop, and a second ctrl-C will force close the client's connection (and therefore cancel everything outstanding on the server).

t0yv0:
I tried reading this but I'm also not sure what behavior we want and would like to understand that before getting down to how it's implemented. If my review's needed here would appreciate a quick sync to figure this out. We have a slightly related grey are for me where bridged providers try to handle graceful termination, and this would be super helpful to clarify what is the desired behavior on handling graceful provider termination, and what is the idiomatic Go expression of that.

Naively I'm expecting something along the lines of the GRPC server handling everything necessary for Contexts that handlers receive, but then us needing to tie a request for graceful termination from the Engine to some for of message to the gRPC server that would instruct it to cancel out all existing context objects and wait for the requests to finish processing. I'd expect this all to be available at this layer. Not at the logical layer as here.

middleware/cancel/cancel.go
```diff
@@ -12,50 +12,49 @@
...
+//
+// Wrapping a provider in `cancel.Wrap` ensures 2 things:
+//
+// 1. When a resource operation times out, the associated context is canceled.
+//
+// 2. When `Cancel` is called, all outstanding gRPC methods have their associated contexts
+// canceled.
+//
+// A `cancel.Wrap`ed provider will still call the `Cancel` method on the underlying
+// provider. If NotImplemented is returned, it will be swallowed.
```
iwahbe, line 26: I added some docs clarifying what `cancel` *currently* does. This behavior is unchanged before and after this PR. The PR only prevents panics on concurrent creations and cancelations.

Since this PR fixes a P1 (panic) and doesn't change the design of `cancel.Wrap`, I suggest that we merge as is unless there are suggestions on how to better implement the existing semantic.

I can tell that there is disagreement on the *desired* behavior of `cancel.Wrap`. I'm happy to have that discussion, but I think it should be a separate discussion, outside the scope of this PR.

CC @blampe @t0yv0

iwahbe: I opened https://github.com/pulumi/pulumi-go-provider/issues/208 to track a design discussion on what `cancel` should do.

blampe:
> Yes. That is the only reason we need to deal with concurrency here.
> 
> Calling `Stop` does cancel the contexts associated with gRPC connects, but it also kills those connections, preventing the provider from communicating any information back to the engine.

Gotcha -- this is the crux of it. You want to cancel the requests but still let them log or return to the engine.

> The [documentation for `Cancel`](https://github.com/pulumi/pulumi/blob/28f098a660aa698c83a77058da28f7867707d0a0/proto/pulumi/provider.proto#L70-L75) says:
> 
> ```proto
>     // Cancel signals the provider to gracefully shut down and abort any ongoing resource operations.
>     // Operations aborted in this way will return an error (e.g., `Update` and `Create` will either return a
>     // creation error or an initialization error). Since Cancel is advisory and non-blocking, it is up
>     // to the host to decide how long to wait after Cancel is called before (e.g.)
>     // hard-closing any gRPC connection.
>     rpc Cancel(google.protobuf.Empty) returns (google.protobuf.Empty) {}
> ```
> 
> To do this, we need a middle ground between `Stop` and `GracefullStop`: we need to communicate to each resource that it should report progress and exit without slamming shut the gRPC connection. AFAIK the convention in go to inform a process that it should clean up and go home is to cancel it's `context.Context`.
>
> The first ctrl-c sends the `Cancel` request to to the provider. The second ctrl-c has the engine kill the provider. As a provider author, we just get the 1 cancel. As I read the description of the `Cancel` call, it seems pretty clear that it _should_ impact ongoing requests.

Interesting, I guess I interpreted this differently. The signal/advisory parts makes it seem like more of a courtesy call, and the "Operations aborted in this way..." explains how the engine will interpret operations aborted by the provider (if it decides to abort anything).

In any case this is probably one of those under-specified corners of behavior. I'm curious what we do in the bridge -- if this is already the norm then I don't have any concerns.

> Naively I'm expecting something along the lines of the GRPC server handling everything necessary for Contexts that handlers receive, but then us needing to tie a request for graceful termination from the Engine to some for of message to the gRPC server that would instruct it to cancel out all existing context objects and wait for the requests to finish processing. I'd expect this all to be available at this layer. Not at the logical layer as here.

Same. This feels like it could be handled by an [interceptor](https://pkg.go.dev/google.golang.org/grpc#UnaryServerInterceptor) which the server can broadcast a shutting-down signal to.

middleware/cancel/cancel.go
blampe: Sounds good to me!

blampe (approved):

PR #210, opened 2024-04-05
Support partial resource initialization errors

iwahbe:
This allows a provider to indicate that it tried to Create, Update or Read a resource but was not able atomically succeed, leaving the resource in a partially initialized state.


iwahbe:
This is motivated by https://github.com/pulumi/pulumi-pulumiservice/issues/84, which requires control over partial initialization.

infer/resource.go
```diff
@@ -905,11 +909,32 @@ func (rc *derivedResourceController[R, I, O]) Create(ctx p.Context, req p.Create
 	}
 
 	id, o, err := (*r).Create(ctx, req.Urn.Name(), input, req.Preview)
-	if err != nil {
+	if initFailed := (ResourceInitFailedError{}); errors.As(err, &initFailed) {
+		defer func(createErr error) {
```
VenelinMartinov, line 913: Why are you deferring the error handling here? Does all the following code allow for a failed create? Is a ResourceInitFailedError Create guaranteed to return a valid id and state?

infer/resource.go
```diff
@@ -926,10 +951,12 @@ func (rc *derivedResourceController[R, I, O]) Create(ctx p.Context, req p.Create
 	return p.CreateResponse{
 		ID:         id,
 		Properties: m,
-	}, nil
+	}, err
```
VenelinMartinov, line 954: How does returning an explicit value interact with the deferred editing of the named returns? Which one wins?

infer/resource.go
```diff
@@ -955,9 +982,30 @@ func (rc *derivedResourceController[R, I, O]) Read(ctx p.Context, req p.ReadRequ
 		}, nil
 	}
 	id, inputs, state, err := read.Read(ctx, req.ID, inputs, state)
-	if err != nil {
+	if initFailed := (ResourceInitFailedError{}); errors.As(err, &initFailed) {
+		defer func(readErr error) {
```
VenelinMartinov, line 986: same questions as above.

VenelinMartinov (commented):
Looks sensible but I'm not familiar with pulumi-go-provider.

Does this code run before the provider implementation of Update, Create etc or does it wrap the provider implementation?

iwahbe:
> Does this code run before the provider implementation of Update, Create etc or does it wrap the provider implementation?

Providers based on `infer` doesn't implement global methods for `Update`, `Create`, etc. Instead resources each define their own version of these methods. This code wraps the resource level implementation.

infer/resource.go
iwahbe: By returning an error, it is assigned to `retError`. The `defer`ed function get's to manipulate `retError` (and `resp`) to manipulate the value ultimately returned by `Create`.

infer/resource.go
iwahbe: > Why are you deferring the error handling here? Does all the following code allow for a failed create?

If we get a `ResourceInitFailedError`, we still want to serialize the output and return, but we also want to keep the context from the original error. It does allowed for a fully failed create, in which case the error is just returned.

> Is a ResourceInitFailedError Create guaranteed to return a valid id and state?

A correct implementation will return a valid enough id and state to be handled in by the resource's `Update` method.

VenelinMartinov (approved):

PR #215, opened 2024-04-08
Add support for state migrations in `infer`

### LINKED ISSUE #193: UpgradeResourceState
Currently, if a resource is serialized with an older version of the schema there's no way to recover it short of manually edit the state file.

iwahbe:
Adds support for state migrations in the `infer` provider. Resources opt into state migrations when they implement the `infer.CustomStateMigrations[O]` interface:

```go
type CustomStateMigrations[O any] interface {
	// StateMigrations is the list of know migrations.
	//
	// Each migration should return a valid State object.
	//
	// The first migration to return a non-nil Result will be used.
	StateMigrations(ctx p.Context) []StateMigrationFunc[O]
}

// StateMigrationFunc represents a stateless mapping from an old state shape to a new
// state shape. Each StateMigrationFunc is parameterized by the shape of the type it
// produces, ensuring that all successful migrations end up in a valid state.
//
// To create a StateMigrationFunc, use [StateMigration].
type StateMigrationFunc[New any] interface {} // Sealed

// StateMigration creates a mapping from an old state shape (type Old) to a new state
// shape (type New).
//
// If Old = [resource.PropertyMap], then the migration is always run.
//
// Example:
//
//	type MyResource struct{}
//
//	type MyInput struct{}
//
//	type MyStateV1 struct {
//		SomeInt *int `pulumi:"someInt,optional"`
//	}
//
//	type MyStateV2 struct {
//		AString string `pulumi:"aString"`
//		AInt    *int   `pulumi:"aInt,optional"`
//	}
//
//	func migrateFromV1(ctx p.Context, v1 StateV1) (infer.MigrationResult[MigrateStateV2], error) {
//		return infer.MigrationResult[MigrateStateV2]{
//			Result: &MigrateStateV2{
//				AString: "default-string", // Add a new required field
//				AInt: v1.SomeInt, // Rename an existing field
//			},
//		}, nil
//	}
//
//	// Associate your migration with the resource it encapsulates.
//	func (*MyResource) StateMigrations(p.Context) []infer.StateMigrationFunc[MigrateStateV2] {
//		return []infer.StateMigrationFunc[MigrateStateV2]{
//			infer.StateMigration(migrateFromV1),
//		}
//	}
func StateMigration[Old, New any, F func(p.Context, Old) (MigrationResult[New], error)](f F) StateMigrationFunc[New] { ... }


// MigrationResult represents the result of a migration.
type MigrationResult[T any] struct {
	// Result is the result of the migration.
	//
	// If Result is nil, then the migration is considered to have been unnecessary.
	//
	// If Result is non-nil, then the migration is considered to have completed and
	// the new value state value will be *Result.
	Result *T
}
```

This allows each resource to define a list of (possibly typed) migrations from old to new state. This design makes the common case of migrating from StateV1 to StateV2 fully typed, but permits an untyped to typed escape hatch.

Fixes #193

### DIFF
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -35,8 +35,8 @@ import (
 	"github.com/pulumi/pulumi-go-provider/middleware/schema"
 )
 
-// A resource that understands how to create itself. This is the minimum requirement for
-// defining a new custom resource.
+// A [custom resource](https://www.pulumi.com/docs/concepts/resources/) inferred from code. This is the
+// minimum requirement for defining a new custom resource.
 //
 // This interface should be implemented by the resource controller, with `I` the resource
 // inputs and `O` the full set of resource fields. It is recommended that `O` is a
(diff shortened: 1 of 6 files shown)


iwahbe:
This is required for https://github.com/pulumi/pulumi-pulumiservice/issues/84, since pulumi-pulumiservice already maintains it's own state migrations.

infer/resource.go
```diff
@@ -932,16 +1047,38 @@ func (rc *derivedResourceController[R, I, O]) Create(ctx p.Context, req p.Create
...
+	if enc, s, err := hydrateFromState[R, I, O](ctx, req.Properties); err == nil {
+		stateEncoder = enc
+		state = s
+	} else {
+		// That didn't work, so maybe we can get by decoding without state migration but by tolerating
+		// missing fields.
+		stateEncoder, err = ende.DecodeTolerateMissing(req.Properties, &state)
+		if err != nil {
+			return p.ReadResponse{}, err
+		}
```
blampe, line 1079: Good fix 👍 I think I ran into this a few times

infer/tests/migrate_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+	return infer.MigrationResult[MigrateStateV2]{
+		Result: &MigrateStateV2{
+			AString: m["aString"].StringValue(),
+			AInt:    int(m["aInt"].NumberValue()),
+		},
+	}, nil
+}
+
+func migrateFromV0(ctx p.Context, v0 MigrateStateV0) (infer.MigrationResult[MigrateStateV2], error) {
```
blampe, line 65: These tests are really helpful to see how this is used in practice.

blampe (approved):
Looks good, let's see how it goes!

PR #222, opened 2024-04-16
[infer] Delegate Configure to the wrapped provider

### LINKED ISSUE #221: Infer does not delegate `Configure` to wrapped providers
### What happened?

I noticed that `infer` based providers do not delegate `Configure` calls to wrapped providers. To support https://github.com/pulumi/pulumi-go-provider/issues/216, they should.

### Example

NA

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

iwahbe:
Fixes #221

Delegate `Configure` calls to an underlying provider if any. Right now, this PR only delegates `Configure` calls, not `DiffConfig` or `CheckConfig`.

### DIFF
--- a/infer/provider.go
+++ b/infer/provider.go
@@ -25,6 +25,8 @@ import (
 	"github.com/pulumi/pulumi-go-provider/middleware/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 type configKeyType struct{}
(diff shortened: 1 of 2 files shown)


infer/provider.go
```diff
@@ -140,7 +142,21 @@ func Wrap(provider p.Provider, opts Options) p.Provider {
 
 	config := opts.Config
 	if config != nil {
-		provider.Configure = config.configure
+		if prev := provider.Configure; prev != nil {
+			provider.Configure = func(ctx p.Context, req p.ConfigureRequest) error {
+				err := config.configure(ctx, req)
```
thomas11, line 147: Since we're calling both `Configure`'s in this case, it's not so much delegation as wrapping, isn't it?

Wouldn't it be more common to call the outermost layer first, then the more specific one last?

iwahbe: > Since we're calling both `Configure`'s in this case, it's not so much delegation as wrapping, isn't it?

Yeah, it's a wrapping call. Delegate was a bad choice of PR name. The function this is in is actually called `Wrap`.

> Wouldn't it be more common to call the outermost layer first, then the more specific one last?

That is what is happening. `config.configure` is the inferred "outer" layer and `prev` is the "inner" layer.

thomas11 (approved):

PR #228, opened 2024-04-24
Match rpc behavior

iwahbe:
While working on https://github.com/pulumi/pulumi-pulumiservice/pull/258, I noticed that native providers receive unknown values over the wire regardless of configure options. This reflects that in the `rpc` middleware.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -146,15 +146,28 @@ type DiffResponse struct {
 	DetailedDiff map[string]PropertyDiff
 }
 
+type diffChanges bool
+
+func (c diffChanges) rpc() rpc.DiffResponse_DiffChanges {
+	if c {
+		return rpc.DiffResponse_DIFF_SOME
+	}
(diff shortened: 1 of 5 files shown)


provider.go
```diff
@@ -151,7 +151,7 @@ func (d DiffResponse) rpc() *rpc.DiffResponse {
 		DeleteBeforeReplace: d.DeleteBeforeReplace,
 		Changes:             rpc.DiffResponse_DIFF_NONE,
 		DetailedDiff:        detailedDiff(d.DetailedDiff).rpc(),
-		HasDetailedDiff:     true,
+		HasDetailedDiff:     len(d.DetailedDiff) == 0,
```
thomas11, line 154: `> 0`?

iwahbe: 🤦 Thanks for the catch. This will need a test before merging.

thomas11 (approved):

PR #229, opened 2024-04-24
Assert that `infer.Wrap`ed providers can access the infer config

iwahbe:
This just adds a test case.

### DIFF
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -47,7 +47,9 @@ func TestInferConfigWrap(t *testing.T) {
 	var baseConfigureWasCalled bool
 	var inferConfigureWasCalled bool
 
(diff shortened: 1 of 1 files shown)


tests/config_test.go
```diff
@@ -58,14 +60,25 @@ func TestInferConfigWrap(t *testing.T) {
 				baseConfigureWasCalled = true
 				return nil
 			},
+			Check: func(ctx context.Context, _ p.CheckRequest) (p.CheckResponse, error) {
+				infer.GetConfig[*testConfig](ctx) // Will panic if config is missing
```
thomas11, line 64: It would be clearer for the next person to have an assertion failure instead of a panic.

thomas11 (approved):

PR #231, opened 2024-04-25
Remove integration go.mod

blampe:
> A module is a collection of packages that are released, versioned, and distributed together.

There's no need to version this subdirectory separately from the repo's root module because they are tightly coupled.

* The last time we remembered to tag this module was `integration/v0.10.0`, which is old enough to have incompatibilities with `main`. A new tag `integration/v0.16.0` is needed for compatibility with `v0.16.0`, so these are not versioned separately.

* The invalid `github.com/pulumi/pulumi-go-provider v0.0.0-00010101000000-000000000000` constraint makes the module un-importable on its own, it _must_ be imported alongside `github.com/pulumi/pulumi-go-provider` -- so we're already distributing these together.
> go: downloading github.com/pulumi/pulumi-go-provider v0.0.0-00010101000000-000000000000
go: github.com/blampe/gomodexample imports
	github.com/pulumi/pulumi-go-provider/integration imports
	github.com/pulumi/pulumi-go-provider: github.com/pulumi/pulumi-go-provider@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000 

* The `replace github.com/pulumi/pulumi-go-provider => ../` directive is another smell that this module is actually just a member of the parent module, since it wants the parent module's code _at the same commit_.

In other words, we don't release/version/distribute these things separately, so there's no need to treat them as separate modules. We can make life easier for ourselves by removing this `go.mod`.

The only impact this will have on downstream is that they will need to remove `github.com/pulumi/pulumi-go-provider/integration` from `go.mod`, since it will become an ambiguous import alongside `github.com/pulumi/pulumi-go-provider`.


integration/integration.go
```diff
@@ -60,7 +60,7 @@ type server struct {
 	context context.Context
 }
 
-func (s *server) ctx(urn presource.URN) context.Context {
+func (s *server) ctx(_ presource.URN) context.Context {
```
iwahbe, line 63:
```suggestion
func (s *server) ctx(presource.URN) context.Context {
```

iwahbe (approved):
I'm vaguely uncomfortable with sharing dev dependencies with prod dependencies, but Go is not.

blampe:
> I'm vaguely uncomfortable with sharing dev dependencies with prod dependencies, but Go is not.

At the end of the day everything is just a package, and yeah it's normal to group together test helpers and mocks into something like [httptest](https://pkg.go.dev/net/http/httptest).

For what it's worth, the dependencies you take on by importing this from the "pulumi-go-provider" module versus importing it from a separate "pulumi-go-provider/integration" module are identical. In other words, in both cases you only take on the subset of dependencies that  `integration/go.mod` currently shows.

I think this has bitten us in pu/pu because our helpers (`pkg/v3/testing/integration` in particular) are heavy in the sense that the end up pulling in a large chunk of business logic.

PR #233, opened 2024-04-26
Minimal example showing that Assets have no signatures

thomas11:
Running this example using an [Asset](https://www.pulumi.com/docs/concepts/assets-archives/) fails with
```
Diagnostics:
  asset:index:A (ourAsset):
    error: unrecognized signature '' in property map for "localAsset"
```

As established with @Frassle, it's because the Asset's signature is lost after RegisterResource.

```
I0426 08:22:01.482624     634 rpc.go:77] Marshaling property for RPC[Provider[command, 0x140019a4480].Check(urn:pulumi:dev::command-ec2-remote::command:remote:CopyFile::copy).news]: 4dabf18193072939515e22adb298388d={c44067f5952c0a294b673a41bacd8c17}
...
I0426 08:22:01.484267     634 rpc.go:292] Unmarshaling property for RPC[Provider[command, 0x140019a4480].Check(urn:pulumi:dev::command-ec2-remote::command:remote:CopyFile::copy).inputs]: 4dabf18193072939515e22adb298388d={}
```


iwahbe:
Thanks for the awesome starting point! I root caused the issue to a bug in pulumi/pkg. Luckily the fix is easy:

```patch
 sdk/go/common/resource/asset/asset.go | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/sdk/go/common/resource/asset/asset.go b/sdk/go/common/resource/asset/asset.go
index 5f0069b97..04f36e90a 100644
--- a/sdk/go/common/resource/asset/asset.go
+++ b/sdk/go/common/resource/asset/asset.go
@@ -213,7 +213,7 @@ func Deserialize(obj map[string]interface{}) (*Asset, bool, error) {
 		uri = u
 	}
 
-	return &Asset{Hash: hash, Text: text, Path: path, URI: uri}, true, nil
+	return &Asset{Sig: AssetSig, Hash: hash, Text: text, Path: path, URI: uri}, true, nil
 }
 
 // HasContents indicates whether or not an asset's contents can be read.
```

I'll open a PR shortly.

PR #236, opened 2024-05-03
Show that assets work as expected

iwahbe:
This locks in the behavior fix from https://github.com/pulumi/pulumi/pull/16073.


tests/grpc/asset_test.go
```diff
@@ -0,0 +1,87 @@
...
+type A struct{}
+
+type AssetInputs struct {
+	LocalAsset *resource.Asset `pulumi:"localAsset,optional"`
+}
+
+type AssetState struct{}
+
+func (*A) Create(ctx context.Context, name string, input AssetInputs, preview bool) (id string, output AssetState, err error) {
+	panic("FOO")
```
thomas11, line 84: ?

iwahbe: Create is never called here, but I should change the panic message here.

thomas11 (approved):

PR #239, opened 2024-05-24
DefaultCheck should apply default values

### LINKED ISSUE #238: Defaults are not applied by `DefaultCheck`
### What happened?

As part of investigating an internal irregularity, I saw that defaults don't get applied by `infer.DefaultCheck`. Since the purpose of `infer.DefaultCheck` is to mirror the default behavior of the `Check` call, defaults should be applied.

It is very clear from the implementation of `DefaultCheck` that defaults are not applied:

https://github.com/pulumi/pulumi-go-provider/blob/327ff50024bcaf4bdb5233cb87ceb0cafb8a5bb7/infer/resource.go#L894-L903

Likewise, it is clear where defaults are applied in non-standard `Check`:

https://github.com/pulumi/pulumi-go-provider/blob/327ff50024bcaf4bdb5233cb87ceb0cafb8a5bb7/infer/resource.go#L872-L876

To fix this issue, we should rewrite the latter half of `Check` in terms of `infer.DefaultCheck` to avoid such discrepancies in the future.

### Example

NA

### Output of `pulumi about`

NA

### Additional context

_No response_

### Contributing

thomas11:
Currently, defaults get applied by non-standard `Check` but not by `infer.DefaultCheck`. Since the purpose of `infer.DefaultCheck` is to mirror the default behavior of the Check call, defaults should be applied.

Resolves #238

### DIFF
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -895,6 +895,9 @@ func (rc *derivedResourceController[R, I, O]) Check(ctx context.Context, req p.C
 func DefaultCheck[I any](inputs resource.PropertyMap) (I, []p.CheckFailure, error) {
 	_, i, err := ende.Decode[I](inputs)
 	if err == nil {
+		if err := applyDefaults(&i); err != nil {
+			return i, nil, fmt.Errorf("unable to apply defaults: %w", err)
+		}
 		return i, nil, nil
 	}
 
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -851,8 +851,18 @@ func (*derivedResourceController[R, I, O]) getInstance() *R {
 }
 
 func (rc *derivedResourceController[R, I, O]) Check(ctx context.Context, req p.CheckRequest) (p.CheckResponse, error) {
+	encoder, i, failures, err := decodeCheckingMapErrors[I](req.News)
+	if err != nil {
+		return p.CheckResponse{}, err
+	}
+	if len(failures) > 0 {
+		return p.CheckResponse{
+			Inputs:   req.News,
+			Failures: failures,
+		}, nil
+	}
+
 	var r R
-	encoder, i, err := ende.Decode[I](req.News)
 	if r, ok := ((interface{})(r)).(CustomCheck[I]); ok {
 		// The user implemented check manually, so call that
 		i, failures, err := r.Check(ctx, req.Urn.Name(), req.Olds, req.News)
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):
This looks reasonable to me ... a little tricky to follow how each of these functions are called and how they might interact with each other, but seems like they're just parallel implementation from your experimentation.

Looks like this just replicates the behaviour from:
https://github.com/pulumi/pulumi-go-provider/blob/1a6b4f88f49c574e5c321ecb4b12e1abe50124ea/infer/resource.go#L872-L874

Would it be possible to make `derivedResourceController.Check` call into this method so we inherit the behaviour rather than mirror the implementation (no idea if this would be feasible and I think this tactical fix is good for an initial change - just thinking for future consideration)?

infer/resource.go
```diff
@@ -895,6 +895,9 @@ func (rc *derivedResourceController[R, I, O]) Check(ctx context.Context, req p.C
 func DefaultCheck[I any](inputs resource.PropertyMap) (I, []p.CheckFailure, error) {
```
iwahbe, line 895: Can we improve the doc comment here?

Suggestion:
> DefaultCheck performs the same validation that leaving `Check` unimplemented does.
> 
> It verifies that inputs serialize cleanly into T and applies defaults as necessary.

infer/resource.go
```diff
@@ -895,6 +895,9 @@ func (rc *derivedResourceController[R, I, O]) Check(ctx context.Context, req p.C
...
 	_, i, err := ende.Decode[I](inputs)
 	if err == nil {
+		if err := applyDefaults(&i); err != nil {
```
iwahbe, line 898: Ideally, I'd now like to see a change to how `Check` is implemented to use `DefaultCheck` instead of duplicating the logic. This is what I meant by:

> To fix this issue, we should rewrite the latter half of Check in terms of infer.DefaultCheck to avoid such discrepancies in the future.

infer/resource_test.go
```diff
@@ -449,3 +449,20 @@ func TestHydrateFromState(t *testing.T) {
...
+
+type inputForDefaultCheck struct {
+	P1 string `pulumi:"str,optional"`
+}
+
+func (i *inputForDefaultCheck) Annotate(a Annotator) {
+	a.SetDefault(&i.P1, "default")
+}
+
+func TestDefaultCheckAppliesDefaultValues(t *testing.T) {
```
iwahbe, line 461: This is a good unit test. It would be great if we could parameterize by if `Check` if implemented (and just returns `DefaultCheck`) vs left as default. We could then assert that they behave identically.

We can leave that to a future exercise.

iwahbe (approved):
A couple suggestions (very similar to @danielrbradley). The change looks good.

infer/resource_test.go
thomas11: I implemented this in the latest commit. Sharing `DefaultCheck` required separating decoding from applying defaults, and didn't really resulted in any LOC savings, but at least the logic is now guaranteed to be the same.

infer/resource.go
thomas11: I implemented this in the latest commit. Sharing DefaultCheck required separating decoding from applying defaults, and didn't really resulted in any LOC savings, but at least the logic is now guaranteed to be the same.

iwahbe (approved):

PR #242, opened 2024-06-04
Introduce AssetOrArchive to support the SDK's Asset type which can be both

### LINKED ISSUE #237: Asset properties can be Archives at runtime
When defining a property of type `resource.Asset`, the generated schema will contain a `$ref` of `pulumi.json#/Asset`. However, in the SDKs, an "asset" is represented by `pulumi.AssetOrArchive` because an archive is currently always a valid asset.

Therefore, when deserializing the PropertyValue for the provider to use, it's valid for the user to have specified one of the Archive types which then isn't deserializable into the `resource.Asset` field within the provider.

1. We need instead a new `resource.AssetOrArchive` type to be used with the go-provider.
2. We need to handle reflecting over `resource.AssetOrArchive`
3. Using `resource.Asset` in a provider field should be considered an error. Though initially this should be a warning to avoid a breaking change but should be upgraded to an error on the next major release.

thomas11:
Resolves #237.

This PR addresses an issue with the design of the [assets and archives support](https://www.pulumi.com/docs/concepts/assets-archives/) in Pulumi core. Quoting #237:

> When defining a property of type resource.Asset, the generated schema will contain a $ref of pulumi.json#/Asset. However, in the SDKs, an "asset" is represented by pulumi.AssetOrArchive because an archive is currently always a valid asset.

> Therefore, when deserializing the PropertyValue for the provider to use, it's valid for the user to have specified one of the Archive types which then isn't deserializable into the resource.Asset field within the provider.

To work around this, we define a new type `types.AssetOrArchive` here in pulumi-go-provider. In the schema, it's represented as a standard `pulumi.json#/Asset` for interoperability. An incoming core `Asset` is translated into a `types.AssetOrArchive`, in `Decode`, with either Asset or Archive populated. Vice versa, `Encode` pulls the asset/archive out of AssetOrArchive.

## Notes
- The decode and encode implementations are bespoke and there might be opportunities to move it to a new, more generic extension point.


infer/internal/ende/ende.go
```diff
@@ -188,8 +189,17 @@ func (e *ende) walk(
...
 			}
-		// This is a scalar value, so we can return it as is.
+		// This is a scalar value, so we can return it as is. The exception is types.AssetOrArchive, which we translate
+		// to a types.Asset or types.Archive, depending on what it contains, to match the SDK's AssetOrArchive type.
 		default:
+			if typ == reflect.TypeOf(types.AssetOrArchive{}) {
+				if v.IsAsset() {
+					v = resource.NewPropertyValue(types.AssetOrArchive{Asset: v.AssetValue()})
+				} else if v.IsArchive() {
+					v = resource.NewPropertyValue(types.AssetOrArchive{Archive: v.ArchiveValue()})
```
iwahbe, line 199: I'm amazed that this works, but if tests show that it does...

iwahbe (commented):
The serde stuff looks good to me. This is the kind of place that I'd really appreciate integration tests to lock in good behavior.

infer/internal/ende/ende.go
thomas11: Could you elaborate on what you expected?

iwahbe: I'm not sure what I expected here. Probably an end run around pulumi/pulumi's mapper.

infer/types/asset.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+
+type AssetOrArchive struct {
+	Asset   *resource.Asset   `pulumi:"asset,optional"`
+	Archive *resource.Archive `pulumi:"archive,optional"`
```
iwahbe, line 21: I thought we discussed using a special sig here to ensure we found `AssetOrArchive` and not some other type.
```suggestion
	Asset   *resource.Asset   `pulumi:"somereallylongsigforassets,optional"`
	Archive *resource.Archive `pulumi:"somereallylongsigforarchives,optional"`
```

infer/internal/ende/ende.go
```diff
@@ -293,16 +309,52 @@ func (e *ende) walkMap(
...
+
+	// If we see "asset" or "archive" properties who have the magic signatures inside, we assume we have an
+	// AssetOrArchive and need to pull the asset or archive out of the object.
+	// TODO,tkappler This is not safe: a user could have defined a property that looks like an
+	// AssetOrArchive but isn't.
+	m := resource.NewPropertyValueRepl(props,
+		nil, // keys are not changed
+		func(a any) (resource.PropertyValue, bool) {
+			if aMap, ok := a.(map[string]any); ok {
+				if rawAsset, ok := aMap["asset"]; ok {
```
iwahbe, line 335: Instead of calling it `"asset"`, lets use a sig:
```suggestion
				if rawAsset, ok := aMap["somereallylongsigforassets"]; ok {
```

This way we can be pretty sure that if we see the key, it's what we are looking for.

infer/resource.go
```diff
@@ -1248,6 +1248,7 @@ func (rc *derivedResourceController[R, I, O]) Delete(ctx context.Context, req p.
 		if err != nil {
 			return err
 		}
+
```
iwahbe, line 1251: Can we not adjust whitespace in files we don't otherwise touch?
```suggestion
```

infer/internal/ende/ende.go
```diff
@@ -300,9 +315,38 @@ func (e *ende) Encode(src any) (resource.PropertyMap, mapper.MappingError) {
...
+	// AssetOrArchive. See #237 for more background.
+	// The literal magic signatures are from pulumi/pulumi and are not exported by the SDK.
+	m := resource.NewPropertyValueRepl(props,
+		nil, // keys are not changed
+		func(a any) (resource.PropertyValue, bool) {
+			if aMap, ok := a.(map[string]any); ok {
+				if rawAsset, ok := aMap[types.AssetSignature]; ok {
+					if asset, ok := rawAsset.(map[string]any); ok {
+						if sig, ok := asset["4dabf18193072939515e22adb298388d"]; ok {
+							if sigStr, ok := sig.(string); ok && sigStr == "c44067f5952c0a294b673a41bacd8c17" {
```
iwahbe, line 330: Can we use the [constant in `pulumi/pulumi`](https://github.com/pulumi/pulumi/blob/5a83cfe6e4b77705240499134fae29d68c3c3ddf/sdk/go/common/resource/sig/sig.go#L18-L21)?
```suggestion
						if kind, ok := asset[sig.Key]; ok {
							if kind, ok := kind.(string); ok && kind == sig.AssetSig {

```

infer/internal/ende/ende.go
```diff
@@ -300,9 +315,38 @@ func (e *ende) Encode(src any) (resource.PropertyMap, mapper.MappingError) {
...
-		resource.NewPropertyMapFromMap(props),
-	)
+
+	// If we see the magic signatures meaning "asset" or "archive", it's an AssetOrArchive and need
+	// to pull the actual, inner asset or archive out of the object and discard the outer
+	// AssetOrArchive. See #237 for more background.
+	// The literal magic signatures are from pulumi/pulumi and are not exported by the SDK.
+	m := resource.NewPropertyValueRepl(props,
+		nil, // keys are not changed
+		func(a any) (resource.PropertyValue, bool) {
```
iwahbe, line 325: Can we make this a top level function to avoid the nesting?

infer/types/asset.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+
+const AssetSignature = "a9e28acb8ab501f883219e7c9f624fb6"
+const ArchiveSignature = "195f3948f6769324d4661e1e245f3a4d"
+
+type AssetOrArchive struct {
+	Asset   *resource.Asset   `pulumi:"a9e28acb8ab501f883219e7c9f624fb6,optional"`
+	Archive *resource.Archive `pulumi:"195f3948f6769324d4661e1e245f3a4d,optional"`
+}
```
iwahbe, line 25: Can we add a test that we will enforce a required `AssstOrArchive`? I want to make sure that this is safe:

```go
type MyArgs struct {
   Input types.AssetOrArchive `pulumi:"input"`
}

func (...) Create(...) ... {
    if args.Asset == nil && args.Archive == nil && !preview {
        panic("THIS SHOULD NEVER BE POSSIBLE")
    }
}
```

infer/internal/ende/ende.go
```diff
@@ -300,9 +315,38 @@ func (e *ende) Encode(src any) (resource.PropertyMap, mapper.MappingError) {
...
+			if aMap, ok := a.(map[string]any); ok {
+				if rawAsset, ok := aMap[types.AssetSignature]; ok {
+					if asset, ok := rawAsset.(map[string]any); ok {
+						if sig, ok := asset["4dabf18193072939515e22adb298388d"]; ok {
```
iwahbe, line 329: Let's make sure we error (or panic) if both the asset and archive keys are set.

infer/types/asset.go
iwahbe, line 25: This needs a doc comment explaining what it is and why you should use it. It *must* specify that this represents a union and setting both `Asset` and `Archive` will fail to serialize.

infer/resource.go
thomas11: I had touched and apparently left this artifact when I reversed my changes.

infer/internal/ende/ende.go
```diff
@@ -338,7 +361,38 @@ const (
...
+		rawArchive, hasArchive := aMap[types.ArchiveSignature]
+
+		if hasAsset && hasArchive {
+			panic(`Encountered both an asset and an archive in the same AssetOrArchive. This
+should never happen. Please file an issue at https://github.com/pulumi/pulumi-go-provider/issues.`)
+		}
+
+		raw := rawAsset
+		if hasArchive {
+			raw = rawArchive
```
danielrbradley, line 376: This line is missing test coverage

infer/internal/ende/ende.go
```diff
@@ -338,7 +361,38 @@ const (
...
+				if kind, ok := kind.(string); ok {
+					if kind == sig.AssetSig || kind == sig.ArchiveSig {
+						// It's an asset/archive inside an AssetOrArchive. Pull it out.
+						return resource.NewObjectProperty(resource.NewPropertyMapFromMap(asset)), true
+					}
+					panic(`Encountered an unknown kind in an AssetOrArchive. This should never
+happen. Please file an issue at https://github.com/pulumi/pulumi-go-provider/issues.`)
+				}
+			}
+		}
```
danielrbradley, line 390: These lines have no unit test coverage - probably worth adding.

infer/schema.go
```diff
@@ -127,6 +128,11 @@ func serializeTypeAsPropertyType(
 			Ref: "pulumi.json#/Archive",
 		}, nil
 	}
+	if t == reflect.TypeOf(types.AssetOrArchive{}) {
+		return schema.TypeSpec{
+			Ref: "pulumi.json#/Asset",
+		}, nil
+	}
```
danielrbradley, line 135: Should we add a warning here if they're using the previous Asset-only type above as it doesn't round-trip and they should migrate to AssetOrArchive?

danielrbradley (commented):
This looks pretty close. I'm happy the implementation looks correct, just a few last bits of polish.

infer/internal/ende/ende.go
```diff
@@ -338,7 +361,38 @@ const (
 	isEmptyArr = iota
 )
 
-// Mark a encoder as generating values only.
+func flattenAssets(a any) (resource.PropertyValue, bool) {
```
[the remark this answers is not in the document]
danielrbradley: Here's an example of a function we could extract to make this handling easier:

```go
func extractAssetOrArchive(aMap map[string]any, typeSig, kindSig string) (map[string]any, bool) {
	insideType, hasTypeSig := aMap[kindSig]
	if !hasTypeSig {
		return nil, false
	}
	insideTypeMap, ok := insideType.(map[string]any)
	if !ok {
		return nil, false
	}
	kind, hasKind := insideTypeMap[sig.Key]
	if !hasKind {
		return nil, false
	}
	kindStr, ok := kind.(string)
	if !ok {
		return nil, false
	}
	if kindStr != typeSig {
		return nil, false
	}
	return insideTypeMap, true
}
```

We can then be more specific on how we want to handle each error case too - panic or error etc.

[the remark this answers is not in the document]
danielrbradley: The method above then simplifies to:

```go
	if aMap, ok := a.(map[string]any); ok {
		if asAsset, isAsset := extractAssetOrArchive(aMap, sig.AssetSig, types.AssetSignature); isAsset {
			return resource.NewObjectProperty(resource.NewPropertyMapFromMap(asAsset)), true
		}
		if asArchive, isArchive := extractAssetOrArchive(aMap, sig.ArchiveSig, types.ArchiveSignature); isArchive {
			return resource.NewObjectProperty(resource.NewPropertyMapFromMap(asArchive)), true
		}
	}
````

File: infer/internal/ende/ende.go [hunk 13]
```diff
@@ -338,7 +361,38 @@ const (
...
+	if aMap, ok := a.(map[string]any); ok {
+		rawAsset, hasAsset := aMap[types.AssetSignature]
+		rawArchive, hasArchive := aMap[types.ArchiveSignature]
+
+		if hasAsset && hasArchive {
+			panic(`Encountered both an asset and an archive in the same AssetOrArchive. This
+should never happen. Please file an issue at https://github.com/pulumi/pulumi-go-provider/issues.`)
```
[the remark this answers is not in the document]
iwahbe: I believe that this is when we are serializing. We need to serialize `AssetOrArchive` into either an asset or an archive, and we can't do that. We can't compress `struct { Asset, Archive }` into an `Asset` or `Archive` without losing information. It is *always* a provider bug and right now `panic`ing is a reasonable option. It would be better to return an error, but it's not clear how.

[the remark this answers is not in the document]
iwahbe: Platform will not create a new type that has our sigs (we assume that sigs won't conflict by accident).

infer/schema.go
thomas11: Screenshot 2024-06-11 at 20 39 46

infer/internal/ende/ende.go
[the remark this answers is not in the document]
thomas11: I don't quite see the simplification here TBH. I've commented and slightly reorganized the method in the latest commit, take a look if that looks better to you.

infer/internal/ende/ende.go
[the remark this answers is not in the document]
danielrbradley: It was the other way round - when we're parsing - we're adding a panic if we recieve something which contains both asset or archive information. My suggestion here would be to just check if it's an asset first, then check if it's an archive. The code below would already do that if the panic removed, if I read it correctly, though it prefers archive over asset.

infer/internal/ende/ende.go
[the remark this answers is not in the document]
danielrbradley: It's certainly partially subjective, so I wouldn't block on this.

Some observations of what makes it look complex to me are:
1. 5 levels of nested if statements means that the error handling paths are somewhat implicit rather than explicit - falling through to the null return (unless it ends up at the panic). The `extractAssetOrArchive` style would give the option for returning specific error messages for each case if we needed them too easier than adding `else` blocks for each `if` level. 
2. The panic "Encountered an unknown kind in an AssetOrArchive" is not adjacent to the condition that would trigger it - making the reader track back to work out the context. I.e. it's inverse of the branch it didn't take 4 lines above (`kind == sig.AssetSig || kind == sig.ArchiveSig`).
3. It's tricky follow if we're looking at an asset or we're looking at an archive, though it then appears it kinda doesn't matter after line 765 when we set the `raw` variable.

infer/internal/ende/ende.go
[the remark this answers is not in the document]
iwahbe: We should panic on ambiguity, as it is always a provider or framework bug. The user will never see the ambiguity and the engine can only pass ambiguous information if it copies the sigs we use, which it should not.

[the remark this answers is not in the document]
danielrbradley: I agree if it's user or framework error, but this data is coming from the engine – for which we don't have a super-good specification of what they might send, so I'd consider being defensive over making contract assertions. But perhaps it's reasonable here to fail hard because it would be a weird change engine side that _hopefully_ they wouldn't make because it's likely not backward compatible (not that we define backward compatibility in our protocols either 😞).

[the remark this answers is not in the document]
iwahbe: There is **no** remotely backwards compatible change (beyond copying the sigs we created so they wouldn't be copied, on purpose or by accident) that would allow this to be an engine bug. It will **always** be a provider author or framework bug, and it should panic to allow the responsible party to fix it, instead of doing the wrong thing for end users.

infer/internal/ende/ende.go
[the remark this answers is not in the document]
thomas11: The paths leading to the default return are not error paths. It's entirely expected that we encounter maps that are not AssetOrArchive, and fall through.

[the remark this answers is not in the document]
danielrbradley: There's also implicit error paths too:
- If it's got the asset type sig, but it's not a map within
- If it's an asset type sig, but has no kind field
- The kind field is not a string

These all just return null, which I think it fine, but it's implicit.

infer/schema.go
thomas11: Hmm, I just noticed that this also causes:
```
$ pulumi package get-schema /Users/tkappler/pulumi/pulumi-command/bin/pulumi-resource-command
warning: Please use pulumi-go-provider's types.AssetOrArchive type instead of resource.Asset or resource.Archive. ...
```

The reason is that types are serialized recursively into the schema, so first `AssetOrArchive`, then its inner `Asset` and `Archive` types.

infer/internal/ende/ende.go
```diff
@@ -300,9 +312,15 @@ func (e *ende) Encode(src any) (resource.PropertyMap, mapper.MappingError) {
...
 		return nil, err
 	}
-	m := resource.NewObjectProperty(
-		resource.NewPropertyMapFromMap(props),
-	)
+
+	// If we see the magic signatures meaning "asset" or "archive", it's an AssetOrArchive and need
+	// to pull the actual, inner asset or archive out of the object and discard the outer
+	// AssetOrArchive. See #237 for more background.
+	// The literal magic signatures are from pulumi/pulumi and are not exported by the SDK.
```
iwahbe, line 319: This is not true, and we no longer have literal signautes.

infer/schema.go
iwahbe: Let's leave a warning as a follow up PR.

infer/types/asset.go
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package types
+
+import "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+
+const AssetSignature = "a9e28acb8ab501f883219e7c9f624fb6"
+const ArchiveSignature = "195f3948f6769324d4661e1e245f3a4d"
```
iwahbe, line 20: Let's try to add a docstring to each exposed part of the API.

```suggestion
const (
    // The signature for the asset part of [AssetOrArchive]
    AssetSignature = "a9e28acb8ab501f883219e7c9f624fb6"
    // The signature for the archive part of [AssetOrArchive]
    ArchiveSignature = "195f3948f6769324d4661e1e245f3a4d"
)
```

Even better, could we move these to `ende` and keep them private. I don't want them to show up in autocomplete since users should never need to use them.

infer/types/asset.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2022, Pulumi Corporation.
```
iwahbe, line 1:
```suggestion
// Copyright 2024, Pulumi Corporation.
```

iwahbe (approved):
This looks good to me. There are a couple simple changes to make before merging, but it mostly looks good to go.

infer/schema.go
thomas11: https://github.com/pulumi/pulumi-go-provider/issues/243

infer/types/asset.go
thomas11: Ah, good call. In contrast to the type, the constants are only used in ende.

PR #244, opened 2024-06-13
AssetOrArchive shouldn't be in the schema

thomas11:
Before this exclusion, the schemas of providers using `AssetOrArchive` would have its type definition in their schema. Besides being wrong anyway, it would also generate invalid Go code because the signatures fields start with numbers.

### DIFF
--- a/infer/tests/types_test.go
+++ b/infer/tests/types_test.go
@@ -1,0 +1,76 @@
+// Copyright 2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


danielrbradley (approved):

danielrbradley (commented):
We could perhaps do with some tests around the schema generation part for these kinds of cases.

iwahbe (commented):
Let's make sure we have a schema test here before merging. I would add the test in `infer/tests` and have it call the GetSchema method on a provider that has an `resource.Asset`, `resource.Archive` and `types.AssetOrArchive`.

thomas11:
> Let's make sure we have a schema test here before merging. I would add the test in `infer/tests` and have it call the GetSchema method on a provider that has an `resource.Asset`, `resource.Archive` and `types.AssetOrArchive`.

Done.

iwahbe (approved):

PR #245, opened 2024-06-21
Extend `infer` with annotations for aliases and resource deprecations

thomas11:
This allows authors of providers built on `infer` to [alias](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#alias) and [deprecate](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#resource) resources.

### DIFF
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -280,6 +280,10 @@ type Annotator interface {
 	//	mypkg:mymodule:MyResource
 	//
 	SetToken(module, name string)
+
+	AddAlias(module, name string)
+
+	SetResourceDeprecationMessage(message string)
 }
 
 // Annotated is used to describe the fields of an object or a resource. Annotated can be
--- a/internal/introspect/annotator.go
+++ b/internal/introspect/annotator.go
@@ -32,10 +32,12 @@ func NewAnnotator(resource any) Annotator {
 
 // Implements the Annotator interface as defined in resource/resource.go
 type Annotator struct {
-	Descriptions map[string]string
-	Defaults     map[string]any
-	DefaultEnvs  map[string][]string
-	Token        string
+	Descriptions       map[string]string
+	Defaults           map[string]any
+	DefaultEnvs        map[string][]string
+	Token              string
+	Aliases            []string
+	DeprecationMessage string
 
 	matcher FieldMatcher
 }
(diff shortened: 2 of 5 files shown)


danielrbradley (approved):

flostadler (approved):
Nice!

infer/resource.go
```diff
@@ -280,6 +280,10 @@ type Annotator interface {
 	//	mypkg:mymodule:MyResource
 	//
 	SetToken(module, name string)
+
+	AddAlias(module, name string)
```
lunaris, line 284: No doc comments and examples?

infer/resource.go
```diff
@@ -280,6 +280,10 @@ type Annotator interface {
...
+
+	SetResourceDeprecationMessage(message string)
```
lunaris, line 286: Nitpicking, but given that this is backed in the implementation by a field called `DeprecationMessage`, I'm inclined to think `Resource` is superfluous here? Moreover, if we're feeling cheeky, we could call it `Deprecate` to align with `Describe`.

danielrbradley: Yeah, good thoughts .. perhaps also `AddAlias` could just be `Alias`?

thomas11: I called it `SetResourceDeprecationMessage ` because resource properties can also be deprecated. The field `DeprecationMessage` is on the resource level so there's no ambiguity.

For the aliases, I went with "add" because it can be called more than once to add different aliases, which I wanted to call attention to.

Let me know if my reasoning doesn't make sense to you.

infer/resource.go
thomas11: Oopsie, that was an oversight.

internal/introspect/annotator.go
```diff
@@ -88,11 +90,25 @@ func (a *Annotator) SetDefault(i any, defaultValue any, env ...string) {
...
+// Panics when module or token are invalid.
+func formatToken(module, token string) string {
 	if !tokens.IsQName(module) {
 		panic(fmt.Sprintf("Module (%q) must comply with %s, but does not", module, tokens.QNameRegexp))
 	}
 	if !tokens.IsName(token) {
 		panic(fmt.Sprintf("Token (%q) must comply with %s, but does not", token, tokens.NameRegexp))
 	}
-	a.Token = fmt.Sprintf("pkg:%s:%s", module, token)
+	return fmt.Sprintf("pkg:%s:%s", module, token)
```
thomas11, line 113: @iwahbe I left the hard-coded `pkg:` as-is, but is that correct? Shouldn't it be whatever package name the provider author has defined?

infer/resource.go
iwahbe: `Describe` also applies to both the resource and it's fields. I agree that `Deprecate` would have been a more consistent name.

internal/introspect/annotator.go
iwahbe: It's correct. A provider doesn't know it's package name until runtime, so we leave in a dummy name: `pkg`.

infer/resource.go
thomas11: Added in #246

PR #246, opened 2024-06-21
Doc strings for the new methods AddAlias and SetResourceDeprecationMessage

thomas11:
A follow-up PR to #245.

### DIFF
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -281,8 +281,16 @@ type Annotator interface {
 	//
 	SetToken(module, name string)
 
+	// Add an [alias](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#alias) for
+	// this resource.
+	// The given module and name need to be valid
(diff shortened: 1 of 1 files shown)


danielrbradley (approved):

infer/resource.go
```diff
@@ -281,8 +281,16 @@ type Annotator interface {
...
 	SetToken(module, name string)
 
+	// Add an [alias](https://www.pulumi.com/docs/using-pulumi/pulumi-packages/schema/#alias) for
+	// this resource.
+	// The given module and name need to be valid
+	// [QNames](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v3/go/common/tokens#QName)
+	// as defined by the Pulumi SDK.
+	// The module and the name will be assembled into a type specifier of the form `pkg:module:name`,
+	// in the same way `SetToken` does.
 	AddAlias(module, name string)
```
iwahbe, line 291: If this is true, why don't we encode that in the type system? We could make `module` a `tokens.ModuleName` and `name` a `tokens.TypeName`.

Issue #212, opened 2024-04-07
[infer] Richer value representation for custom resources


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Many infer functions operate on strongly typed values, but there is currently no way for these values to encode that a location is secret. We should enhance the surface area of types to allow for this richer information.

Related: https://github.com/pulumi/pulumi-go-provider/pull/155

iwahbe:
If we wanted to be really fancy (and a bit magical), we could implement pointer based lookup functions for input structs:

```go
func IsSecret[T any](p.Context, *T) bool
func IsComputed[T any](p.Context, *T) bool
```

Working similarly to `infer.Annotor`, these functions would be used to index into a precomputed metadata table on the input type by value location. This would require that inputs are passed by reference to preserve location, and would return `false` for any moved value.

iwahbe:
This is a breaking change, but I'm removing 1.0.0-blocker since I don't think we have the time to address this issue before we release 1.0.0. We can continue to iterate and release an `infer/v2`, to be made default in 2.0.0.

PR #253, opened 2024-07-24
`DefaultCheck` with `context.Context`

iwahbe:
Following up on #252, this allows `DefaultCheck` to safely and accurately handle secrets application. The new signature for `DefaultCheck` is easier to extend without additional breaking changes. The downside to this design is that this makes `DefaultCheck` special. I think it's worth living with that until #212 is resolved.

I'd like to merge shortly after #252 (and before #252 is released) so that user's don't rely on #252 injecting secrets when `CustomCheck` is implemented and `DefaultCheck` is not called.

### DIFF
--- a/infer/configuration.go
+++ b/infer/configuration.go
@@ -93,9 +93,13 @@ func (c *config[T]) checkConfig(ctx context.Context, req p.CheckRequest) (p.Chec
 		if req.Urn != "" {
 			name = req.Urn.Name()
 		}
-		i, failures, err := t.Check(ctx, name, req.Olds, req.News)
+		defCheckEnc, i, failures, err := callCustomCheck(ctx, t, name, req.Olds, req.News)
 		if err != nil {
 			return p.CheckResponse{}, err
+		}
+
+		if defCheckEnc != nil {
+			encoder = *defCheckEnc
 		}
 
 		inputs, err := encoder.Encode(i)
@@ -103,7 +107,7 @@ func (c *config[T]) checkConfig(ctx context.Context, req p.CheckRequest) (p.Chec
 			return p.CheckResponse{}, err
 		}
 		return p.CheckResponse{
-			Inputs:   applySecrets[T](inputs),
+			Inputs:   inputs,
 			Failures: failures,
 		}, nil
 	}
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -149,12 +149,9 @@ func TestInferCheckConfigSecrets(t *testing.T) {
 }
 
 type config struct {
-	Field  string `pulumi:"field" provider:"secret"`
-	Nested struct {
-		Int       int    `pulumi:"int" provider:"secret"`
-		NotSecret string `pulumi:"not-nested"`
-	} `pulumi:"nested"`
-	NotSecret string `pulumi:"not"`
+	Field         string `pulumi:"field" provider:"secret"`
+	NotSecret     string `pulumi:"not"`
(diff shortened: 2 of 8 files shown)


infer/configuration.go
```diff
@@ -93,17 +93,23 @@ func (c *config[T]) checkConfig(ctx context.Context, req p.CheckRequest) (p.Chec
...
+		ctx = context.WithValue(ctx, defaultCheckEncoderKey{}, defaultCheckEncoder)
 		i, failures, err := t.Check(ctx, name, req.Olds, req.News)
 		if err != nil {
 			return p.CheckResponse{}, err
 		}
 
+		if defaultCheckEncoder.enc != nil {
+			encoder = *defaultCheckEncoder.enc
+		}
+
```
mjeffryes, line 106: I could use some comments here to understand what's going on. My best guess is:
- DefaultCheck uses the context to return an encoder that obscures secrets along with any errors it returns
- We use this encoder instead of the more general one created earlier so we don't have to apply secrets in here directly.

tests/config_test.go
```diff
@@ -140,33 +137,63 @@ func (c *config) Check(
...
-			"field": resource.NewProperty("value"),
-			"nested": resource.NewProperty(resource.PropertyMap{
-				"int":        resource.NewProperty(1.0),
-				"not-nested": resource.NewProperty("not-secret"),
-			}),
-			"not": resource.NewProperty("not-secret"),
-		},
+	}))
+
+	t.Run("with-defaults", func(t *testing.T) {
```
mjeffryes, line 166: really this is "with-DefaultCheck" right? we're applying defaults either way, but in the one case it's manual in the custom Check method and in the other, it's relying on DefaultCheck (with the only actual difference being that DefaultCheck applies secrets).

mjeffryes (commented):
I'm not sure I follow the motivation. Why do we want authors to apply secrets themselves rather than relying on the framework to apply them?

infer/configuration.go
iwahbe: I extracted this out into a function and added some comments.

iwahbe:
> I'm not sure I follow the motivation. Why do we want authors to apply secrets themselves rather than relying on the framework to apply them?

I added this to the commit message:

> Ideally, most users will not need to implement check. By design, not implementing
`CustomCheck` has the same behavior as implementing check with
`return infer.DefaultCheck(...)`. If users do implement `CustomCheck`, then we don't want
to make any assumptions about what behavior they want. This includes not applying defaults
and not injecting secrets.

EronWright (commented):
It is a breaking change (to `DefaultCheck`), right? Is that a concern?

The technique of using the context to pass information back up, I've not seen elsewhere.  Maybe could use more commentary, like why the default check's encoder is so precious.

iwahbe:
> It is a breaking change (to `DefaultCheck`), right? Is that a concern?

It is. I think its a positive one though. This library is still in `v0` and we tell users to expect breaking changes.

> The technique of using the context to pass information back up, I've not seen elsewhere. Maybe could use more commentary, like why the default check's encoder is so precious.

`ende.Encoder`s are the byproduct of going from `resource.PropertyMap` to `C` where `C` is some user supplied type. They carry a look-aside table of what is expressible with `resource.PropertyMap` but not expressible with `C`, such as secrets and unknown values. I'll add a comment to `ende.Encoder` to explain why it exists.

EronWright (approved):
Approving because it seems fine.

PR #263, opened 2024-08-20
Add support for Call

iwahbe:
This PR adds support for the `Call` gRPC method at the top level `Provider` struct.


t0yv0:
What's the test plan here?

iwahbe:
> What's the test plan here?

I'm trying to determine the correct behavior before adding tests. We will definitely need tests before merging.

iwahbe:
Tests have been added. I think this is ready to merge.

resourcex/traverse.go
```diff
@@ -73,3 +73,47 @@ func Traverse(v resource.PropertyValue, p resource.PropertyPath, f func(resource
...
+
+// Walk traverses a property value along all paths, performing a depth first search.
+func Walk(v resource.PropertyValue, f func(resource.PropertyValue, WalkState)) {
+	walk(v, f, WalkState{})
+}
+
+type WalkState struct {
+	// We are entering the value. The children of the value have not yet been visited.
+	Entering bool
+	// The value is not transitively computed.
```
t0yv0, line 85: This is a bit weird and does not mesh with my mental model:

```
type Value = Unknown
                    | Known ConcreteValue
```

Are there really transitively unknown values or this is just a degenerate encoding artifact that does not occur in practice?

provider.go
```diff
@@ -718,8 +728,78 @@ func (p *provider) StreamInvoke(*rpc.InvokeRequest, rpc.ResourceProvider_StreamI
...
+		resourcex.Walk(v, func(v presource.PropertyValue, state resourcex.WalkState) {
+			if state.Entering || !v.IsOutput() {
+				return
+			}
+			for _, dep := range v.OutputValue().Dependencies {
+				urns = append(urns, string(dep))
+			}
+		})
+
+		returnDependencies[string(name)] = &rpc.CallResponse_ReturnDependencies{Urns: urns}
```
t0yv0, line 778: Is this needed even when the provider has first-class outputs in Return, where this approximation can be un-abmiguously computed? Probably no harm in including but I thought maybe engine does this?

t0yv0:
Nit: I would rename the PR to drop "at the provider level", looks like this introduces support for the Call gRPC method which is great.

t0yv0:
One important question is instance lifetime. When a client issues a series of Call methods against a resource with the same URN, is there a possibility for the provider author to rely on the retention of in-memory instance state across the series of calls? 

```typescript
const c = new Counter()
c.increment()
c.increment()
export const n = c.curentValue /// 2
```

To achieve the above UX would it be possible for the implementation to be equally simple?

t0yv0 (approved):
Code looking good but I had a few non-blocking questions. I also would suggest considering an E2E test involving the real Pulumi CLI and a sample program, at least one of those  for the happy path.

t0yv0:
Corner cases to think of (perhaps handled already in this codebase).

1. Does Call issue during preview. If so how are unknowns handled?

2. If there are schema-level or first-class secrets, does the implementation guarantee reasonable secret wrapping.

resourcex/traverse.go
iwahbe: I'm guessing it's a degenerate value, but lots of serde stuff requires valid types, even on computed fields so it might come up. This is mostly to be cautious.

provider.go
```diff
@@ -718,8 +728,78 @@ func (p *provider) StreamInvoke(*rpc.InvokeRequest, rpc.ResourceProvider_StreamI
...
+	}
+
+	return &rpc.CallResponse{
+		Return:             _return,
+		ReturnDependencies: returnDependencies,
+		Failures:           checkFailureList(resp.Failures).rpc(),
+	}, nil
+}
+
+type CallRequest struct {
```
thomas11, line 793: document public type?

provider.go
```diff
@@ -718,8 +728,78 @@ func (p *provider) StreamInvoke(*rpc.InvokeRequest, rpc.ResourceProvider_StreamI
...
+		MonitorAddr:       req.GetMonitorEndpoint(),
+		Organization:      req.GetOrganization(),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to build pulumi.Context: %w", err)
+	}
+
+	args, err := p.getMap(req.GetArgs())
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert args into a property map")
```
thomas11, line 754: [nit] include the original error to give an indication why conversion failed?

provider.go
```diff
@@ -342,7 +343,11 @@ type Provider struct {
 	Read   func(context.Context, ReadRequest) (ReadResponse, error)
 	Update func(context.Context, UpdateRequest) (UpdateResponse, error)
 	Delete func(context.Context, DeleteRequest) error
-	// TODO Call
+
+	// Call allows methods to be attached to resources.
+	//
+	// Right now, Call is restricted to methods on component resources.
```
thomas11, line 349: Is the last sentence still accurate? If so, maybe give readers a hint why that is.

provider.go
iwahbe: I'm not sure, but there is no documentation saying that it isn't necessary.

It seems like it is not necessary in Pulumi:

https://github.com/pulumi/pulumi/blob/cbfea508e7b2a4f1dd0985cefbeda3a6e891f7f9/pkg/resource/deploy/source_eval.go#L1140-L1142

I'm hesitant to assume that it is never necessary.

iwahbe:
> Corner cases to think of (perhaps handled already in this codebase).
> 
> 1. Does Call issue during preview. If so how are unknowns handled?
> 2. If there are schema-level or first-class secrets, does the implementation guarantee reasonable secret wrapping.

1. I believe that it does. Call is supposed to result in a Pulumi program running (with `pulumi.Output`), so it unknowns should just flow through.
2. Not at all. This implementation just brings Call support into the base level of the library. Semantic guarentees like (if secret in the schema, then secret on the wire) are handled at a higher level (Call is not hooked up to `infer`).

provider.go
iwahbe: I linked the associated issues in pulumi/pulumi.

iwahbe:
> One important question is instance lifetime. When a client issues a series of Call methods against a resource with the same URN, is there a possibility for the provider author to rely on the retention of in-memory instance state across the series of calls?
> 
> ```ts
> const c = new Counter()
> c.increment()
> c.increment()
> export const n = c.curentValue /// 2
> ```
> 
> To achieve the above UX would it be possible for the implementation to be equally simple?

I believe that this is possible now, though I would discourage relying on it. AFAIK the protocol does not make any guarantees about number of provider instances, or the order of calls. In theory, both `c.increment()`s could be called in parallel.

provider.go
iwahbe: I opened https://github.com/pulumi/pulumi/pull/17113 to pin down an answer here.

PR #269, opened 2024-09-06
Interpret JSON Config values

### LINKED ISSUE #171: Automatically unmarshal complex configuration types

iwahbe:
This PR adds a middleware to handle decoding JSON encoded config values. The middleware is marked as deprecated, since we would like to remove this functionality as soon as https://github.com/pulumi/pulumi/pull/15032 closes.

Fixes #171

### DIFF
--- a/middleware/complexconfig/provider.go
+++ b/middleware/complexconfig/provider.go
@@ -1,0 +1,227 @@
+// Copyright 2024, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package complexconfig adds middleware for schema informed complex configuration
+// encoding/decoding as a work-around for https://github.com/pulumi/pulumi/pull/15032.
+//
+// The entry point for this package is [Wrap].
+//
+// Deprecated: This package will be removed after
+// https://github.com/pulumi/pulumi/pull/15032 merges.
+package complexconfig
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/pulumi/pulumi-go-provider/internal/putil"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/sig"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+
+	p "github.com/pulumi/pulumi-go-provider"
+)
+
+func Wrap(provider p.Provider) p.Provider {
+	encoder := provider
+	contract.Assertf(provider.GetSchema != nil, "provider.GetSchema must be implemented")
+	encoder.CheckConfig = encodeCheckConfig(provider.CheckConfig, provider.GetSchema)
+	return encoder
+}
+
+type (
+	getSchema   = func(context.Context, p.GetSchemaRequest) (p.GetSchemaResponse, error)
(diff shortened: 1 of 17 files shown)


middleware/complexconfig/provider.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if putil.IsSecret(v) {
+		defer func() {
+			out = putil.MakeSecret(v)
+		}()
+		v = putil.MakePublic(v)
+	}
+
+	if !v.IsString() || spec.Type == "string" {
```
thomas11, line 103: This seems like a bug at first glance - if it's _not_ a string or if it _is_ a string. But I think I get it now: if the value is not a string, we can't unmarshal it. If the value is string-typed in the spec, there's no point unmarshaling it since it's a primitive string.

thomas11 (approved):

middleware/complexconfig/provider.go
iwahbe: Yes! I see how that could be confusing though. I'll add ☝️ as a comment.

middleware/complexconfig/provider.go
```diff
@@ -0,0 +1,181 @@
...
+				b, ok := v.(bool)
+				return ok && b
+			}
+
+			elem, hasElem := m["value"]
+
+			return resource.NewProperty(resource.Output{
+				Known:        hasElem,
+				Secret:       castBool("secret"),
+				Dependencies: nil,
```
t0yv0, line 171: Always nil?

t0yv0:
Does this need to cover old config in DiffConfig or CheckConfig suffices? Perhaps as a follow-up.

t0yv0 (approved):

middleware/complexconfig/provider.go
iwahbe: No. Good call. The rapid test property values didn't generate interesting dependency values. I addressed this and added valid support for dependency values in a subsequent push.

iwahbe:
> Does this need to cover old config in DiffConfig or CheckConfig suffices? Perhaps as a follow-up.

I think it should be ok 🤞:

From `provider.proto`:

```proto
message DiffRequest {
    ...

    // the new input values of resource to diff, copied from CheckResponse.inputs.
    google.protobuf.Struct news = 4;

    ...
}
```

[src](https://github.com/pulumi/pulumi/blob/770eb1e5450c1e88d0d4ec921b643b9bd565dc74/proto/pulumi/provider.proto#L269-L270)

I'll double check that this applies for providers as well.

iwahbe:
Check config is definitely called on new provider config before diff is called, so I think this should work by just overriding CheckConfig.

PR #272, opened 2024-09-17
Apply secrets to function result properties marked as such

### LINKED ISSUE #259: Add ability to easily mark output of an invoke as secret
## Hello!

## Issue details

I would like to be able to mark the output of an invoke as always secret similar to how it can be done today for a resource being created with `provider:"secret"`. I expect this is likely because the outputs of the functions are plain strings and not `pulumi.Output`

mikhailshilkov:
Function output properties can be annotated as secrets. That may have two effects:

1. The property is marked as a secret in Pulumi schema (already works in pulumi-go-provider). As of today, this will have no effect on Pulumi SDK and the result will still be returned to the engine as plain text.
2. The provider marks the output property value as a secret and that's what gets sent over gRPC to the client SDK.

This PR implements (2) by applying secrets to inferred function results.

Added an invoke test that verifies this behavior. Also, extended the `credentials` sample program with an implementation of a secretty invokes and verified that it works manually.

Note that this requires 3.131.0 or later of Node.js/Go/YAML SDK or a later version for Python SDK support. Earlier versions fail to deserialize a secretty invoke output.

Resolves #259


mikhailshilkov:
This will need a pulumi-yaml update before the tests go green

iwahbe (approved):

iwahbe (changes requested):
For this to be safe, I think it will need an engine level feature flag: The engine tells the provider that it's ok to send secrets on invokes.

Otherwise this will break existing providers on old SDKs without any explanation.

iwahbe:
> 3. A provider upgrades to this new release + pu/pu version it requires, re-runs the codegen to take advantage of the new SDKs.


Will the newly generated SDKs require version's of their languages core SDKs that support this?

mikhailshilkov:
> Will the newly generated SDKs require version's of their languages core SDKs that support this?

I think so... E.g. Node implementation relies on new functions in the core SDK: https://github.com/pulumi/pulumi/pull/17237/files#diff-f786ce189e2b1bb74d5f9425d102b713287faea09bc04b692ea0d42e1b035a10

Besides, taking advantage of output secretness requires a code change from the provider's author, hopefully they will do so only if they know what they do.

iwahbe:
> > Will the newly generated SDKs require version's of their languages core SDKs that support this?
> 
> I think so... E.g. Node implementation relies on new functions in the core SDK: https://github.com/pulumi/pulumi/pull/17237/files#diff-f786ce189e2b1bb74d5f9425d102b713287faea09bc04b692ea0d42e1b035a10
> 
> Besides, taking advantage of output secretness requires a code change from the provider's author, hopefully they will do so only if they know what they do.

I don't see any required version bump in that diff. On `master` it's still on `3.42.0`:

https://github.com/pulumi/pulumi/blob/189559fa8e43bb88b3f4e0ae9ca3093c74103a2b/pkg/codegen/nodejs/gen.go#L49-L50

iwahbe (approved):
Thinking about it some more, the only case where this introduces a break is:
1. Someone will upgrade their pulumi-go-provider version to v${next}, which includes outputty functions.
2. Someone will generate an SDK with an old version of pulumi, which doesn’t require the new output-supporting SDKs.
3. The provider will start sending outputs in invokes, which will break existing usage.

This requires a `pulumi package gen-sdk` version that's behind what `pulumi-go-provider` uses, which is very hard to guard against.

PR #275, opened 2024-09-19
Prevent directly nested output failure for TestRapidDeepEqual

### LINKED ISSUE #274: Workflow failure: Run Acceptance Tests on main
## Workflow Failure

[Run Acceptance Tests on main](https://github.com/pulumi/pulumi-go-provider/blob/main/.github/workflows/main.yaml) has failed. See the list of failures below:

- [2024-09-18T10:06:40.000Z](https://github.com/pulumi/pulumi-go-provider/actions/runs/10919935053)

iwahbe:
Fixes #274

Issue #274 revealed a bug in the test, but not in the actual `DeepEquals` implementation.


internal/putil/putil.go
```diff
@@ -106,6 +109,8 @@ func MakeKnown(v resource.PropertyValue) resource.PropertyValue {
 // 2. It doesn't panic when computed values are present.
```
thomas11, line 109: This comment is not true anymore, is it? `contract.Assert` panics.

thomas11: On second thought, the whole method doesn't seem to make sense anymore. When secrets and computed values are not allowed, it's just vanilla DeepEquals, isn't it?

iwahbe: > This comment is not true anymore, is it? contract.Assert panics.

The comment is still true. `foldOutputValue` normalizes both secret (`.IsSecret()`) and computed (`.IsComputed()`) values into outputs (`.IsOutput()`) values to simplify the comparison. 

👇 is still true (and does not panic):

```go
putil.DeepEquals(
    resource.MakeComputed(resource.MakeSecret(resource.NewProperty(true))),
    resource.MakeSecret(resource.MakeComputed(resource.NewProperty(true))),
)
```

`foldOutputValue` never returns top level `resource.Secret` or `resource.Computed` values, so the asserts are valid.

PR #278, opened 2024-09-24
Extend code coverage to be more meaningful

mikhailshilkov:
I noticed two issues with existing code coverage stats:

1. Only in-module data are collected, e.g. tests in the `github.com/pulumi/pulumi-go-provider/tests` module don't output coverage data for `github.com/pulumi/pulumi-go-provider`
2. The `examples` and `tests` folders have coverage data for its resource implementation code.

This PR changes the arguments to collect data for all submodules of `github.com/pulumi/pulumi-go-provider` but to exclude `examples` and `tests/grpc` in codecov itself.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -6,3 +6,6 @@ coverage:
     patch:
       default:
         informational: true
+ignore:
+  # Examples implementation code should not count against coverage.
+  - "examples/"
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: build build_examples install_examples lint lint-copyright lint-golang
 
-GO_TEST_FLAGS=-race -coverprofile="coverage.txt"
+GO_TEST_FLAGS=-race -coverprofile="coverage.txt" -coverpkg=github.com/pulumi/pulumi-go-provider/...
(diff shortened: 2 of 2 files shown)


codecov.yml
```diff
@@ -6,3 +6,7 @@ coverage:
     patch:
       default:
         informational: true
+ignore:
+  # Examples implementatio and test code should not count against coverage.
```
iwahbe, line 10: Why shouldn't `tests/grpc` count against coverage? Everything in a testing dir should be part of a test, and thus count against coverage.

```suggestion
  # Examples implementation and test code should not count against coverage.
```

mikhailshilkov: As I understand it, it's a part of tests, not an object being tested. Do we want to see the percentage of lines covered in that folder?

iwahbe: I think it's useful as a sanity check on the tests. If there are tests that have low coverage percentages, its a good indicator that something has gone wrong in testing.

I don't feel strongly either way, but I do think that code coverage on the tests themselves is helpful.

mikhailshilkov: I don't feel strong either - I'll return it to err on the side of more data

iwahbe (approved):

PR #281, opened 2024-10-25
Support parameterization at the gRPC level

iwahbe:
This PR adds basic support for provider parameterization.


provider.go
```diff
@@ -1091,6 +1110,79 @@ func (p *provider) Cancel(ctx context.Context, _ *emptypb.Empty) (*emptypb.Empty
 
 }
 
+type (
+	// ParameterizeRequest configures the provider as parameterized.
```
t0yv0, line 1114: I've been curious so I searched - https://pulumi-developer-docs.readthedocs.io/latest/docs/architecture/providers.html#parameterized-providers does a much better treatment of this then the doc comments, linking to end to end use cases. Perhaps drop a link here and keep comments especially for elucidating how the go-provider specific things help the user accomplish what's needed?

t0yv0 (changes requested):
This needs acceptance tests, otherwise LGTM.

t0yv0 (approved):

PR #286, opened 2025-01-13
Always apply schema level secrets

iwahbe:
This will resolve https://github.com/pulumi/pulumi-docker-build/issues/403 when pulumi-docker-build upgrades to this version.


infer/resource.go
```diff
@@ -933,7 +942,6 @@ func callCustomCheck[T any](
 //
 // It also adds defaults to inputs as necessary, as defined by [Annotator.SetDefault].
 func DefaultCheck[I any](ctx context.Context, inputs resource.PropertyMap) (I, []p.CheckFailure, error) {
-	inputs = applySecrets[I](inputs)
```
t0yv0, line 936: Thoughts on maybe leaving this as is? Doing it twice is no harm? 

Perhaps as an undocumented behavior of DecodeCheck it's ok to remove it though.

infer/configuration.go
```diff
@@ -107,7 +107,7 @@ func (c *config[T]) checkConfig(ctx context.Context, req p.CheckRequest) (p.Chec
 			return p.CheckResponse{}, err
 		}
 		return p.CheckResponse{
-			Inputs:   inputs,
+			Inputs:   applySecrets[T](inputs),
```
t0yv0, line 110: What's with this? GitHub complaining randomly?

```
 Check failure on line 402 in provider.go

 
GitHub Actions
/ test / Test (1.22.x, true)

undefined: CustomCheckNoDefaults
```

infer/resource.go
```diff
@@ -865,41 +865,50 @@ func (*derivedResourceController[R, I, O]) getInstance() *R {
...
-
 	var r R
 	if r, ok := ((interface{})(r)).(CustomCheck[I]); ok {
 		// The user implemented check manually, so call that.
 		//
 		// We do not apply defaults or secrets if the user has implemented Check
 		// themselves. Defaults and secrets are applied by [DefaultCheck].
 
-		defCheckEnc, i, failures, err := callCustomCheck(ctx, r, req.Urn.Name(), req.Olds, req.News)
+		backupEncoder, _, _, _ := decodeCheckingMapErrors[I](req.News)
```
t0yv0, line 875: Are you sure `failures`, `err` need no handling here? 

What if the backupEncoder moves under this `if` block:

```
if encoder == nil {
			encoder = &backupEncoder
}
```

So as to compute it lazily if really needed, and it that case still may handle the error?

t0yv0 (approved):

infer/resource.go
thomas11: +1 - it may be correct now but doesn't look robust

infer/tests/check_test.go
```diff
@@ -257,3 +259,45 @@ func TestCheckDefaultsRecursive(t *testing.T) {
 	}, resp.Inputs)
 
 }
+
+func TestCheckAlwaysAppliesSecrets(t *testing.T) {
```
thomas11, line 263: Maybe it's just me, but I feel like this test would benefit from a high-level description of how the parts fit together. Like how the secret-ness of the input comes in via the annotation on `CustomCheckNoDefaultsArgs.Input`.

infer/resource.go
iwahbe: The caller of `DefaultCheck` only witnesses the effects of `applySecrets[I]` here when they see values that were previously secret show up as secret again (in the returned value of `Check`). 

This doesn't change observable behavior from the user's perspective (except that `applySecrets[I]` covers values added in `CustomCheck.Check`, the goal of the PR).

infer/resource.go
iwahbe: > Are you sure failures, err need no handling here?

I'll explain why in a comment. The tldr is: the backup encoder might not be derived from a shape that matches the shape of `I`, so it might not fit correctly. It will default to just doing a reasonable thing.

Issue #261, opened 2024-08-05
Add Lifecycle Hooks for Opening and Closing Handles During Configuration and Shutdown


zbuchheit:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

It would be beneficial to execute a hook that opens a handle during configuration and closes the handle during shutdown. This would be particularly useful for providers that require a sync or commit action upon closure.

### Affected area/feature

Related: https://github.com/pulumi/pulumi/issues/9780

iwahbe:
I'm labeling as https://github.com/pulumi/pulumi-go-provider/labels/awaiting-upstream because we need a life-cycle hook from Pulumi to make this work reliably.

joeduffy:
@iwahbe Wasn't sure what you meant by this. The need is to have a `Close` or `Shutdown` method on the provider itself that we invoke so that the provider can tidy up any resources, or perform shutdown actions like flushing logs, etc. Is this truly awaiting something upstream, or can that be done in the provider interface itself? Might require a new gRPC call?

iwahbe:
We would need a change in the gRPC protocol so that the provider can be told "please cleanup and exit gracefully". I see two options here:

1. We can add a `Shutdown` method. The engine would be required to call `Shutdown` before shutting down a provider (irrespective of if `Cancel` was called).
2. We can change the semantics of the existing `Cancel` call so that:
  a. `Cancel` is always called before a provider exits
  b. `Cancel` implies that the provider does not need to accept future gRPC connections.


I would advocate for (2), since it keeps a smaller API surface area for provider authors. This SDK already assumes (2b).

rshade:
Can we reschedule this one, now that the upstream has been completed?

iwahbe:
The underlying issue was related to the issue here, but not actually the issue we need implemented. Looking at the logs of a `pulumi up`, we see:

```
...
{"method":"/pulumirpc.ResourceProvider/Handshake", ...
...
{"method":"/pulumirpc.ResourceProvider/GetPluginInfo", ...
{"method":"/pulumirpc.ResourceProvider/CheckConfig", ...
...
{"method":"/pulumirpc.ResourceProvider/Configure", ...
...
{"method":"/pulumirpc.ResourceProvider/Check", ...
...
{"method":"/pulumirpc.ResourceProvider/Create", ...
...
```

[...]

PR #301, opened 2025-02-28
Add a test for canceling resources with a timeout

iwahbe:
When I was answering https://github.com/pulumi/pulumi-go-provider/issues/261#issuecomment-2688957783, I was looking at our implementation of `Cancel`. I noticed there wasn't a test for this edge case, and added one.

### DIFF
--- a/tests/cancel_test.go
+++ b/tests/cancel_test.go
@@ -181,3 +181,46 @@ func TestCancelTimeout(t *testing.T) {
 		assert.ErrorIs(t, err, context.DeadlineExceeded)
 	})
 }
+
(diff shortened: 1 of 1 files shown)


tests/cancel_test.go
```diff
@@ -181,3 +181,46 @@ func TestCancelTimeout(t *testing.T) {
 		assert.ErrorIs(t, err, context.DeadlineExceeded)
 	})
 }
+
+// Test that a `Cancel` call will cancel in-flight operations, even if they already have a
+// timeout associated with them.
+//
+//	Main routine                   Server routine                 Cancel routine
```
thomas11, line 188: A+ comment

thomas11 (approved):

PR #304, opened 2025-03-04
Add an example without any middleware

iwahbe:
This PR adds an example to the `examples` section that doesn't use any middleware.


examples/echo/main.go
```diff
@@ -0,0 +1,136 @@
...
+func main() {
+
+	echoType := tokens.NewTypeToken(tokens.NewModuleToken("echo", "index"), "Echo")
+
+	err := p.RunProvider("echo", "v0.1.0-dev", p.Provider{
+		GetSchema: func(context.Context, p.GetSchemaRequest) (p.GetSchemaResponse, error) {
+			// It is recommended to use [github.com/pulumi/pulumi-go-provider/middleware/schema.Wrap] for
+			// structured schema generation.
+			return p.GetSchemaResponse{Schema: `{
+  "name": "echo",
```
thomas11, line 26: Why not read schema.json here instead? Or remove schema.json, doesn't seem like it's used elsewhere?

iwahbe: Because `schema.json` is generated from this string. This keeps the entire provider contained in this file.

We check in a `schema.json` alongside all examples so that potential authors can see the schema that corresponds to the provider. It isn't used anywhere.

thomas11 (approved):

PR #310, opened 2025-03-21
Simplify the component provider implementation

iwahbe:
Stacked on top of #308, this makes the implementation much simpler.


rquitales (approved):
Thanks

PR #312, opened 2025-03-24
Alternative simple component experience

iwahbe:
This PR shows an alternative implementation of the component package introduced in #308.

This method embraces that components are built on top of infer, leaving room for `infer.DefaultProvider` to support the rest of `infer`'s feature set in time.

This implementation makes it much easier to gradually transition from a more scaffolded "default" experience to an experience where the user assumes direct control.

### DIFF
--- a/examples/component-program/main.go
+++ b/examples/component-program/main.go
@@ -12,27 +12,29 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// package main shows how a simple comoponent provider can be created using existing
+// package main shows how a simple component provider can be created using existing
 // Pulumi programs that contain components.
 package main
 
 import (
-	"github.com/pulumi/pulumi-go-provider/component"
+	"fmt"
+	"os"
+
 	"github.com/pulumi/pulumi-go-provider/examples/component-program/nested"
+	"github.com/pulumi/pulumi-go-provider/infer"
 )
 
 func main() {
-	err := component.ProviderHost(
-		component.WithName("go-components"),
-		component.WithVersion("v0.0.1"),
-		component.WithResources(
-			component.ProgramComponent(
-				component.ComponentFn[RandomComponentArgs, *RandomComponent](NewMyComponent)),
-			component.ProgramComponent(
-				component.ComponentFn[nested.NestedRandomComponentArgs, *nested.NestedRandomComponent](nested.CreateNestedRandomComponent))),
-	)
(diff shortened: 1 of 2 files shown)


examples/component-program/main.go
```diff
@@ -12,27 +12,29 @@
...
-		component.WithResources(
-			component.ProgramComponent(
-				component.ComponentFn[RandomComponentArgs, *RandomComponent](NewMyComponent)),
-			component.ProgramComponent(
-				component.ComponentFn[nested.NestedRandomComponentArgs, *nested.NestedRandomComponent](nested.CreateNestedRandomComponent))),
-	)
+	err := infer.Default().
+		WithName("go-components").
+		WithVersion("v0.0.1").
+		WithComponent(
```
iwahbe, line 31: Nits: `WithComponents`.

examples/component-program/main.go
```diff
@@ -12,27 +12,29 @@
...
+			infer.ProgramComponent(NewMyComponent),
```
iwahbe, line 32: Nit: better name
```suggestion
			infer.ComponentFunction(NewMyComponent),
```

infer/default.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Default creates an inferred provider which applies as many defaults as possible.
+func Default() DefaultProvider {
+	return DefaultProvider{}
+}
+
+type DefaultProvider struct {
+	name, version string
+	components    []InferredComponent
+}
```
mjeffryes, line 33: If we're going to add a builder to `infer` we should make sure it's compatible with the custom resource APIs and that it offers a reasonably smooth progression from the easy defaults to highly customized.

infer/default.go
```diff
@@ -0,0 +1,90 @@
...
+				"csharp": map[string]any{
+					"packageReferences": map[string]any{
+						"Pulumi": "3.*",
+					},
+					"respectSchemaVersion": true,
+				},
+			},
+		},
+		Components: p.components,
+	}
```
mjeffryes, line 87: For graceful degradation, the schema options probably ought to be part of the builder (with this as a default), rather than a one-size-fits-all config in `BuildAndRun`.

mjeffryes (commented):
If we introduce new APIs in `infer`, I think we need to be careful that the generalize to all providers, but after discussing this with @rquitales I'm cautiously optimistic that we can make this work.

PR #315, opened 2025-03-24
Rename `(Annotator).SetResourceDeprecationMessage` to `(Annotator).Deprecate`

### LINKED ISSUE #248: Reconsider the design of `infer.Annotator.SetResourceDeprecationMessage`
## Hello!

## Issue details

`infer.Annotator.SetResourceDeprecationMessage` does not fit in with the rest of the API provided by `infer.Annotator`. To set descriptions (similar to deprecation messages), `infer.Annotator` has a `Describe(subject, message)` method. 

I would recommend that we change `SetResourceDeprecationMessage` to `Deprecate(subject any, message string)` before we 1.0 `pulumi-go-provider`.

See discussion: https://github.com/pulumi/pulumi-go-provider/pull/245#pullrequestreview-2132327302

EronWright:
Based on https://github.com/pulumi/pulumi-go-provider/pull/245#pullrequestreview-2132327302 it looks like we want to rename `SetResourceDeprecationMessage` to `Describe`.

This PR also enables field-level deprecation messages.

Closes #248

### DIFF
--- a/infer/README.md
+++ b/infer/README.md
@@ -160,9 +160,14 @@ func (f *FileState) Annotate(a infer.Annotator) {
 }
 ```
 
-To deprecate a resource or field, annotate it with `Deprecate`.
+To deprecate a resource or field, annotate it with `Deprecate`:
 
-The only mandatory method for a `CustomResource` is `Create`.
+```go
+func (f *FileState) Annotate(a infer.Annotator) {
+	a.Deprecate(&f.OldField, "You should prefer to use NewField instead.")
+```
+
+The only mandatory method for a `CustomResource` is `Create`:
 
 ```go
 func (*File) Create(ctx context.Context, name string, input FileArgs, preview bool) (
--- a/infer/README.md
+++ b/infer/README.md
@@ -160,6 +160,8 @@ func (f *FileState) Annotate(a infer.Annotator) {
 }
 ```
 
+To deprecate a resource or field, annotate it with `Deprecate`.
+
 The only mandatory method for a `CustomResource` is `Create`.
 
 ```go
--- a/infer/resource.go
+++ b/infer/resource.go
@@ -292,8 +292,8 @@ type Annotator interface {
 	// `mypkg:mymodule:MyResource`, in the same way `SetToken` does.
 	AddAlias(module tokens.ModuleName, name tokens.TypeName)
 
-	// Set a deprecation message for the resource, which officially marks it as deprecated.
-	Deprecate(message string)
+	// Set a deprecation message for a struct field, which officially marks it as deprecated.
+	Deprecate(i any, message string)
 }
 
 // Annotated is used to describe the fields of an object or a resource. Annotated can be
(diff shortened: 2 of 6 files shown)


iwahbe (approved):

rquitales (approved):

iwahbe (changes requested):
> Sorry for the drive-by comment, but I'm curious: Is there a way to deprecate individual properties?

Good catch! There is not right now, but we should 100% add it. To avoid breaking our user's more then once, let's pair this PR with a change to support deprecating individual properties:

```patch
-SetResourceDeprecationMessage(message string)
+Deprecate(any i, message string)
```

`Deprecate` should fully mirror the semantics of `Describe` with regards to object/field inference.

EronWright:
Thanks folks for review, I'll incorporate the feedback.

infer/README.md
```diff
@@ -160,6 +160,8 @@ func (f *FileState) Annotate(a infer.Annotator) {
 }
 ```
 
+To deprecate a resource or field, annotate it with `Deprecate`.
```
iwahbe, line 163: I think we should provide an example if we mention `Deprecate` here.

````suggestion
To deprecate a resource or field, annotate it with `Deprecate`:

```go
func (f *FileState) Annotate(a infer.Annotator) {
	a.Deprecate(&f.OldField, "You should prefer to use NewField instead.")
```
````

That said, I don't think that `infer/README.md` needs to document every feature. As long as it's well documented in the reference docs ([here](https://pkg.go.dev/github.com/pulumi/pulumi-go-provider@v0.25.0/infer#Annotator)), we should be ok.

File: infer/resource.go [hunk 2]
```diff
@@ -292,8 +292,8 @@ type Annotator interface {
 	// `mypkg:mymodule:MyResource`, in the same way `SetToken` does.
 	AddAlias(module tokens.ModuleName, name tokens.TypeName)
 
-	// Set a deprecation message for the resource, which officially marks it as deprecated.
-	SetResourceDeprecationMessage(message string)
+	// Set a deprecation message for a struct field, which officially marks it as deprecated.
+	Deprecate(i any, message string)
```
iwahbe, line 296: I'd put an example here.
```suggestion
	// Set a deprecation message for a struct field, which officially marks it as deprecated.
	//	func (*s Struct) Annotated(a Annotator) {
	//		a.Deprecate(&s.Field, "field is deprecated")
	//	}
	//
	// To deprecate a resource, object or function, call Deprecate on the struct itself:
	//
	//	func (*s Struct) Annotated(a Annotator) {
	//		a.Deprecate(&s, "Struct is deprecated")
	//	}
	Deprecate(i any, message string)
```

infer/schema.go
```diff
@@ -57,15 +57,18 @@ func getAnnotated(t reflect.Type) introspect.Annotator {
...
 	}
 
 	ret := introspect.Annotator{
-		Descriptions: map[string]string{},
-		Defaults:     map[string]any{},
-		DefaultEnvs:  map[string][]string{},
+		Descriptions:       map[string]string{},
+		Defaults:           map[string]any{},
+		DefaultEnvs:        map[string][]string{},
+		DeprecationMessage: map[string]string{},
```
iwahbe, line 71: Nit: this should be plural like `Descriptions`, `Defaults` and `DefaultEnvs`.
```suggestion
		DeprecationMessages: map[string]string{},
```

iwahbe (approved):
A couple of small comments to address before merging, but otherwise it looks good to me.

Issue #295, opened 2025-02-15
Opt-in to modern schema flags by default


mjeffryes:
Authors adopting pulumi-go-provider should get the current best practice in their schema flags by default. (eg. "respectSchemaVersion": true, etc) We'll need to do this in such a way that we don't force users to new schema flags when

PR #317, opened 2025-03-25
Enable provider generation from Go-based Pulumi component programs

rquitales:
This PR introduces a builder to easily scaffold a new `infer` based provider. This implementation allows for users to easily create their required custom resources, components or functions and expose them with sane defaults. Since the new builder returns a `infer.Options` struct, users can also make further customizations to the provider settings should they feel restricted by the builder functions provider.

## Functional Changes  

- Added a new builder to create a default provider that sets valid defaults
- Added an example test to demonstrate usage.  
- Updated existing documentation.

## Testing  

- Implemented an end-to-end scenario as an example test.  
- Performed manual validation.
- Added unit tests.

---

## Example

```go
type RandomComponent struct {
	pulumi.ResourceState
	RandomComponentArgs
	Password        pulumi.StringOutput `pulumi:"password"`
}

type RandomComponentArgs struct {
	Length pulumi.IntInput `pulumi:"length"`
}

func NewMyComponent(ctx *pulumi.Context, name string,
		compArgs RandomComponentArgs, opts ...pulumi.ResourceOption) (*RandomComponent, error) {
	// Define your component constructor logic here.
}

func main() {
	err := infer.NewProviderBuilder().
		WithName("go-components").
		WithVersion("v0.0.1").
		WithComponents(
			infer.Component(NewMyComponent),
		).
		BuildAndRun()
}
```

Stacked on: #313
Supersedes: #308, #312 
Closes: #295


infer/default_provider.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// WithVersion sets the provider version.
+func (dp *DefaultProvider) WithVersion(version string) *DefaultProvider {
+	dp.version = version
+	return dp
+}
+
+// Build constructs the provider options based on the current state of the default provider.
+func (dp *DefaultProvider) Build() *Options {
```
iwahbe, line 218: Why does this return a `*Options`? `infer.Provider` takes a `Options`.
```suggestion
func (dp *DefaultProvider) Build() Options {
```

infer/default_provider.go
```diff
@@ -0,0 +1,251 @@
...
+
+	if dp.metadata.LanguageMap == nil {
+		dp.metadata.LanguageMap = defaultMetadata.LanguageMap
+	}
+
+	return dp
+}
+
+// WithResources adds the given custom resources to the provider.
+func (dp *DefaultProvider) WithResources(resources ...InferredResource) *DefaultProvider {
```
iwahbe, line 126: Other's might argue, but I'd prefer that this didn't mutate.
```suggestion
func (dp DefaultProvider) WithResources(resources ...InferredResource) *DefaultProvider {
```

infer/default_provider.go
```diff
@@ -0,0 +1,251 @@
...
+		ModuleMap:  dp.moduleMap,
+	}
+}
+
+// validate checks if the default provider configuration is valid.
+func (dp *DefaultProvider) validate() error {
+	switch {
+	case dp.name == "":
+		return fmt.Errorf("provider name is required")
+	case dp.version == "":
```
iwahbe, line 234: I think providers work now with no version.

infer/default_provider_test.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+type MockResource struct{}
+
+type MockResourceArgs struct{}
+
+type MockResourceState struct{}
+
+//nolint:lll
+func (m MockResource) Create(ctx context.Context, name string, args MockResourceArgs, preview bool) (string, *MockResourceState, error) {
```
iwahbe, line 55: nit:
```suggestion
//nolint:lll
func (m MockResource) Create(
	ctx context.Context, name string, args MockResourceArgs, preview bool,
) (string, *MockResourceState, error) {
```

infer/default_provider.go
```diff
@@ -0,0 +1,251 @@
...
+//	func main() {
+//		err := infer.NewDefaultProvider(nil).
+//			WithName("go-components").
+//			WithVersion("v0.0.1").
+//			WithComponents(
+//				infer.Component(NewMyComponent),
+//			).
+//			BuildAndRun()
+//	}
+func NewDefaultProvider(opts *Options) *DefaultProvider {
```
iwahbe, line 67: We should probably mention that things like the language map may change between versions of pulumi-go-provider.

infer/default_provider.go
mikhailshilkov: Yes, `version` is now optional. We will inject a git tag version if this provider is git-hosted.

infer/default_provider.go
```diff
@@ -0,0 +1,251 @@
...
+	return dp
+}
+
+func (dp *DefaultProvider) WithPluginDownloadURL(pluginDownloadURL string) *DefaultProvider {
+	dp.metadata.PluginDownloadURL = pluginDownloadURL
+	return dp
+}
+
+// WithName sets the provider name.
+func (dp *DefaultProvider) WithName(name string) *DefaultProvider {
```
mikhailshilkov, line 206: Can we please also add `Namespace`? (it's a new recent option that defines the namespace for client SDKs)

infer/default_provider.go
rquitales: Using receivers make method chaining easier for users, and is more Go idiomatic from what I've seen in other repositories. I'm more favorable towards using receivers here.

examples/component-provider/randomComponent.go
```diff
@@ -0,0 +1,54 @@
...
+	RandomComponentArgs
+	Password        pulumi.StringOutput `pulumi:"password"`
+	HardcodedOutput pulumi.StringOutput `pulumi:"hardcodedOutput"`
+}
+
+type RandomComponentArgs struct {
+	Length pulumi.IntInput `pulumi:"length"`
+}
+
+func NewMyComponent(ctx *pulumi.Context, name string, compArgs RandomComponentArgs, opts ...pulumi.ResourceOption) (*RandomComponent, error) {
```
EronWright, line 34: Were you planning to actually use `NewNestedRandomComponent` in here? Where are you demonstrating nesting?

rquitales: This is to demonstrate/test the ability that we can use code that is in a different module than one that lives in main. Perhaps the name may need to be clearer, but the current example is not to test nested components, but rather, nested modules.

 Eg. https://github.com/pulumi/pulumi-go-provider/pull/317/files/472ab11c0fb6bb78b60794776a5e642de869e5d4#diff-0a58ad0bb7ae01516ba413795b3cd9e0a0d651f3ce4dff46bbdb1e33c7fc20f5

infer/default_provider.go
```diff
@@ -0,0 +1,232 @@
...
+				"respectSchemaVersion": true,
+			},
+			"go": map[string]any{
+				"generateResourceContainerTypes": true,
+				"respectSchemaVersion":           true,
+			},
+			"python": map[string]any{
+				"requires": map[string]any{
+					"pulumi": ">=3.0.0,<4.0.0",
+				},
```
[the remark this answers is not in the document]
rquitales: Thanks for the reference. I have verified that SDKgen does indeed inject these minimum versions, so I'll remove from our codebase.

infer/default_provider.go
```diff
@@ -0,0 +1,232 @@
...
+	return dp
+}
+
+func (dp *DefaultProvider) WithPluginDownloadURL(pluginDownloadURL string) *DefaultProvider {
+	dp.metadata.PluginDownloadURL = pluginDownloadURL
+	return dp
+}
+
+// WithName sets the provider name.
+func (dp *DefaultProvider) WithName(name string) *DefaultProvider {
```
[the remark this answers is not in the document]
rquitales: The reason why it's not a function arg is that we plan to make this optional in a follow-up PR once we can infer names when running the provider from source. If the provider is compiled, then a name would need to be provided. Having name be supplied as a function arg makes it more explicit.

infer/default_provider.go
```diff
@@ -0,0 +1,232 @@
...
+	dp.name = name
+	return dp
+}
+
+// WithVersion sets the provider version.
+func (dp *DefaultProvider) WithVersion(version string) *DefaultProvider {
```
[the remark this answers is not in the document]
rquitales: `version` is optional as surfaced from an earlier feedback comment: https://github.com/pulumi/pulumi-go-provider/pull/317#discussion_r2014520873

infer/default_provider.go
```diff
@@ -0,0 +1,232 @@
...
+	return dp
+}
+
+// WithModuleMap adds the given module map to the provider.
+func (dp *DefaultProvider) WithModuleMap(moduleMap map[tokens.ModuleName]tokens.ModuleName) *DefaultProvider {
+	dp.moduleMap = moduleMap
+	return dp
+}
+
+func (dp *DefaultProvider) WithLanguageMap(languageMap map[string]any) *DefaultProvider {
```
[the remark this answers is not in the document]
rquitales: Updated.

infer/default_provider.go
```diff
@@ -0,0 +1,232 @@
...
+//				infer.Component(NewMyComponent),
+//			).
+//			BuildAndRun()
+//	}
+//
+// Please note that the initial defaults provided by this function may change with future releases of
+// this framework. Currently, we are setting the following defaults:
+//
+// - LanguageMap: A map of language-specific metadata that is used to generate the SDKs for the provider.
+func NewDefaultProvider() *DefaultProvider {
```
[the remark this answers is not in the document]
rquitales: I like the suggestion of `infer.NewProviderBuilder`.

If we rename to `infer.NewProviderBuilder`, perhaps we should also rename `Build` to `BuildOptions` and `BuildAndRun` to just `Run`? The rationale being that `BuildOptions` would be a way for users to eject from the builder pattern and obtain the raw Options struct for further processing if necessary.

@mikhailshilkov @mjeffryes WDYT?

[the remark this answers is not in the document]
mjeffryes: +1 on `NewProviderBuilder` and renaming `Build` to `BuildOptions`. Ambivalent on renaming `BuildAndRun`

examples/component-provider/schema.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "go-components",
```
mikhailshilkov, line 2: Is the `namespace` missing? I see it in the test and in my local example, so won't block on this.

mikhailshilkov (approved):
LGTM, thank you all!

PR #326, opened 2025-04-09
fix: update integration test to error if update fails

rquitales:


rquitales:

- #326 ◀

integration/integration.go
```diff
@@ -297,6 +297,10 @@ func (l LifeCycleTest) Run(t *testing.T, server Server) {
 				Olds: olds,
 				News: check.Inputs.Copy(),
 			})
+			if !update.ExpectFailure && err != nil {
+				assert.NoError(t, err, "failed to update the resource")
+				continue
```
EronWright, line 302: It doesn't make much sense to `continue` here, because if the update failed unexpectedly, it is probably nonsensical and simply wasteful to apply any subsequent updates.  Maybe it should `break`, or fail?  Is it supposed to fall down to `Delete`?

Also curious why the preview's `err` is ignored.

rquitales: I'm reusing the same pattern as below to continue to the next loop iteration. This was helpful for debugging a change I introduced in another PR that ended up breaking updates, which is why I decided to separate it into its own PR. While we could consider updating this integration test suite to fail early, the test runs quickly (within seconds), so I'd prefer to prioritize other work items for now.

rquitales: That makes sense. I have updated to `require.NoError` to bail early. Added a check for preview's `err` as well.

rquitales: A `LifeCycleTest` can include multiple update operations, and it would be helpful to know if any others also fail during the same test run. Since these operations complete quickly, I think it's better to run all of them up front rather than requiring the author to re-run tests repeatedly in a whack-a-mole fashion.

Issue #302, opened 2025-03-01
Update infer API to use request/response structs for data passing


rquitales:
Currently, data is passed from provider RPC to `infer` implemented methods as function arguments. This is brittle, as it prevents us from exposing new fields from the framework/RPC without requiring a breaking change from provider authors when they upgrade the framework. Instead, we should adopt a similar model as the lower-level API with a request/response struct for passing data.

An example implementation would be:

```go
package infer

...

type CreateRequest[I] struct {
    Input I
    URN        presource.URN         // the Pulumi URN for this resource.
    Properties presource.PropertyMap // the provider inputs to set during creation.
    Timeout    float64               // the create request timeout represented in seconds.
    Preview    bool 
}

type CreateResponse[O] struct {
     ID string
     Output O
}

type CustomResource[I, O any] interface {
	// All custom resources must be able to be created.
	CustomCreate[I, O]
}

type CustomCreate[I, O any] interface {
	Create(ctx context.Context, req CreateRequest) (resp CreateResponse[O], err error)
}

```

PR #325, opened 2025-04-09
refactor(infer): switch to request/response structs for infer custom resource interface functions

rquitales:
**Description:**  
This PR updates the custom `Create`, `Read`, `Update`, `Delete`, `Diff`, and `Check` function interfaces in the `infer` package to use structured request/response types instead of relying on function arguments and return values.

**Why this matters:**  
By adopting request/response structs, we improve long-term maintainability and extensibility for users of this package. Future enhancements—such as additional input/output fields—can be introduced without breaking changes to function signatures.

⚠️ **Breaking Change**:
This refactor introduces a breaking change for consumers of the `infer` package. Implementations of the `CustomCreate`, `CustomRead`, `CustomUpdate`, `CustomDelete`, `CustomDiff`, and `CustomCheck` interfaces will need to be updated to the new request/response-style signatures.

**What’s included:**  
- Refactored all relevant function signatures to use request/response structs.  
- Updated all usages and associated tests accordingly.  
- Updated `infer` package documentation to reflect the new API style.  
- Code changes are organized in reviewable commits.  
- No behavioral or logic changes are introduced.

**Before vs. After Example:**

**Before:**
```go
type CustomCreate[I, O any] interface {
	Create(ctx context.Context, name string, inputs I, preview bool) (id string, output O, err error)
}
```

**After:**
```go
type CreateRequest[I any] struct {
	Name    string
	Inputs  I
	Preview bool
}

type CreateResponse[O any] struct {
	ID     string
	Output O
}

type CustomCreate[I, O any] interface {
	Create(ctx context.Context, req CreateRequest[I]) (CreateResponse[O], error)
}
```

**Note:**  
This is a structural refactor only. No new behavior or logic is introduced.

Closes: #302

PR #324, opened 2025-04-09
refactor(types): simplify generic type arguments

rquitales:
This PR simplifies generic type arguments across the codebase to improve readability and maintainability.
Note: This change does not introduce any functional changes.

Issue #276, opened 2024-09-23
Rewrite `Construct` handling (middleware) in line with `Call`


iwahbe:
The interface provided by `provider.Provider.Construct` is extremely hard to work with and test against. The interface was "designed" during hack-week and was thus extremely rushed.

https://github.com/pulumi/pulumi-go-provider/blob/c63d6b63a27dfde32bca0f26716207ac195c64ee/provider.go#L1033-L1044

Downsides are:
- Provider authors are never given access to the raw `resource.PropertyMap` inputs and outputs.
- The interface is confusing to use, it's a lambda on a input which must be called to return the output.
- Not compatible with `rpc.Provider` interop: #219
- Not compatible with test mocking: #82

`provider.Provider.Call` has a lot of the same constraints, but a much more reasonable interface:

https://github.com/pulumi/pulumi-go-provider/blob/c63d6b63a27dfde32bca0f26716207ac195c64ee/provider.go#L815-L832

Note that `Call` has the same set of constraints as `Construct`, but there was more time spent doing design work.

To make it easier/possible to support #82, #219 and #266, we should re-write `Construct` in the same style as `Call`.

This will be a breaking change, so it must go in before 1.0.0. This should not affect `infer` based components, since they are authored at a higher level. It will affect how `infer` implements components, and will enable better error messages (or even fixes) to issues like https://github.com/pulumi/pulumi-go-provider/issues/270.

EronWright:
Digging into this, I would guess that one of the challenges during hack-week was related to the `context.Context` parameter that is passed down and eventually baked into the Pulumi context struct.  There's a provider.Wrap function that seeks to replace the context, and the wrapper must be invoked prior to newing up the Pulumi context, else the wrapper's context won't be applied.

iwahbe:
@EronWright Couldn't the wrapper use `(*pulumi.Context).WithValue`?

Commenting on the history, dealing with `context.Context` wasn't the problem during the hack week, when the provider used it's own `infer.Context` struct here.

EronWright:
Good news, I found a reasonable solution (IMO) after a few false starts.  Here's a preview:
https://github.com/pulumi/pulumi-go-provider/compare/main...issue-276-2

The gist of the solution is to move the `Construct` function parameter of `ConstructRequest` into the Go context, to be made available by the provider server to the middleware. This is exposed as a `ProviderHost` interface.

[...]

PR #327, opened 2025-04-09
refactor(infer): switch to request/response structs for call interface

rquitales:
**Description:**  
This PR updates the custom `Call` function interface in the `infer` package to use structured request/response types instead of relying on function arguments and return values.

This change is aligned with #325 and targets the `Function` resource, instead of custom resources.

⚠️ **Breaking Change**:
This refactor introduces a breaking change for consumers of the `infer` package. Implementations of the `CustomCall` interface will need to be updated to the new request/response-style signatures.

**What’s included:**  
- Refactored all relevant function signatures to use request/response structs.  
- Updated all usages and associated tests accordingly.  
- Updated `infer` package documentation to reflect the new API style.  
- Code changes are organized in reviewable commits.  
- No behavioral or logic changes are introduced.

**Before vs. After Example:**

**Before:**
```go
type Fn[I any, O any] interface {
	// A function is a mapping from `I` to `O`.
	Call(ctx context.Context, args I) (output O, err error)
}
```

**After:**
```go
// FunctionRequest wraps the input type for a function call
type FunctionRequest[I any] struct {
	// Input contains the function input arguments.
	Input I
}

// FunctionResponse wraps the output type for a function call
type FunctionResponse[O any] struct {
	// Output contains the function result.
	Output O
}

// Fn is a function (also called fnvoke) inferred from code. `I` is the function input,
// and `O` is the function output. Both must be structs.
type Fn[I any, O any] interface {
	// A function is a mapping from `I` to `O`.
	Call(ctx context.Context, req FunctionRequest[I]) (resp FunctionResponse[O], err error)
}
```

**Note:**  
This is a structural refactor only. No new behavior or logic is introduced.

Closes: #302

### DIFF
--- a/infer/function.go
+++ b/infer/function.go
@@ -29,11 +29,25 @@ import (
 	"github.com/pulumi/pulumi-go-provider/middleware/schema"
 )
 
+// FunctionRequest wraps the input type for a function call
+type FunctionRequest[I any] struct {
+	// Input contains the function input arguments.
+	Input I
+}
+
+// FunctionResponse wraps the output type for a function call
+type FunctionResponse[O any] struct {
+	// Output contains the function result.
+	Output O
+}
+
 // Fn is a function (also called fnvoke) inferred from code. `I` is the function input,
 // and `O` is the function output. Both must be structs.
 type Fn[I any, O any] interface {
-	// A function is a mapping from `I` to `O`.
-	Call(ctx context.Context, input I) (output O, err error)
+	// Fn is a function (also called an "invoke" or "Provider Function") inferred from code. `I` is the function input,
+	// and `O` is the function output. Both must be structs.
+	// See: https://www.pulumi.com/docs/iac/concepts/resources/functions/#provider-functions
+	Call(ctx context.Context, req FunctionRequest[I]) (resp FunctionResponse[O], err error)
 }
 
 // InferredFunction is a function inferred from code. See [Function] for creating a
@@ -145,11 +159,11 @@ func (r *derivedInvokeController[F, I, O]) Invoke(ctx context.Context, req p.Inv
 	if v := reflect.ValueOf(f); v.Kind() == reflect.Pointer && v.IsNil() {
 		f = reflect.New(v.Type().Elem()).Interface().(F)
 	}
-	o, err := f.Call(ctx, i)
+	o, err := f.Call(ctx, FunctionRequest[I]{Input: i})
 	if err != nil {
 		return p.InvokeResponse{}, err
 	}
-	m, err := encoder.Encode(o)
+	m, err := encoder.Encode(o.Output)
 	if err != nil {
 		return p.InvokeResponse{}, err
 	}
--- a/examples/credentials/main.go
+++ b/examples/credentials/main.go
@@ -118,10 +118,12 @@ func (*User) Diff(ctx context.Context, req infer.DiffRequest[UserArgs, UserState
 
 type Sign struct{}
 
-func (Sign) Call(ctx context.Context, args SignArgs) (SignRes, error) {
+func (Sign) Call(ctx context.Context, req infer.FunctionRequest[SignArgs]) (infer.FunctionResponse[SignRes], error) {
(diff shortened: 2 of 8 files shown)


rquitales:

- #324
    - #325
        - #327 ◀

EronWright (changes requested):
Since we're making a breaking change to Call, we should consider exposing the `*pulumi.Context` that is currently being discarded [here](https://github.com/pulumi/pulumi-go-provider/blob/a019455cf196d45f706cace3d9742ece9b90c33a/provider.go#L804).

I would really like to understand why, in [provider.go](https://github.com/pulumi/pulumi/blob/dc8858698b58c5da5e3813f7aecf14939caaed41/sdk/go/pulumi/provider/provider.go#L84-L105), the CallRequest is adapted to the Pulumi Go SDK? Shouldn't the infer library expose the resultant Context?
> // Call adapts the gRPC CallRequest/CallResponse to/from the Pulumi Go SDK programming model.

How this would impact this PR is that the Call method in `infer` would have a `*pulumi.Context` rather than a `context.Context` as first parameter. I'm proposing that same change for `Construct` in https://github.com/pulumi/pulumi-go-provider/issues/276.

image

iwahbe:
@EronWright the `(infer.Fn[I,O]).Call` function being changed here is unrelated to the gRPC Call method. `infer.Fn` represents a function, and does not have access to a `*pulumi.Context`. The Call gRPC method represents a method on a component resource. That Call method is not currently hooked into `infer` at all.

EronWright:
Thanks @iwahbe for helping to clarify, that makes sense.  How would folks feel then about renaming the `(infer.Fn[I,O]).Call`  function to `(infer.Fn[I,O]).Invoke`, to prepare for exposing the `Call` stuff later?

It was a bit confusing because at the go-provider level, we do have support for Invoke and for Call.

https://github.com/pulumi/pulumi-go-provider/blob/a019455cf196d45f706cace3d9742ece9b90c33a/provider.go#L346-L347

https://github.com/pulumi/pulumi-go-provider/blob/a019455cf196d45f706cace3d9742ece9b90c33a/provider.go#L364-L370

infer/function.go
```diff
@@ -29,11 +29,23 @@ import (
...
+	// Output contains the function result.
+	Output O
+}
+
 // Fn is a function (also called fnvoke) inferred from code. `I` is the function input,
 // and `O` is the function output. Both must be structs.
 type Fn[I any, O any] interface {
 	// A function is a mapping from `I` to `O`.
-	Call(ctx context.Context, input I) (output O, err error)
+	Call(ctx context.Context, req FunctionRequest[I]) (resp FunctionResponse[O], err error)
```
EronWright, line 48: I propose we rename this to `Invoke` to greatly reduce confusion vis-a-vis resource methods.

infer/function.go
```diff
@@ -29,11 +29,23 @@ import (
...
+}
+
+// FunctionResponse wraps the output type for a function call
+type FunctionResponse[O any] struct {
+	// Output contains the function result.
+	Output O
+}
+
 // Fn is a function (also called fnvoke) inferred from code. `I` is the function input,
 // and `O` is the function output. Both must be structs.
```
EronWright, line 45:
```suggestion
// Fn is a function (also called an "invoke" or "Provider Function") inferred from code. `I` is the function input,
// and `O` is the function output. Both must be structs.
// see: https://www.pulumi.com/docs/iac/concepts/resources/functions/#provider-functions
```

infer/function.go
rquitales: I've split the `Call` → `Invoke` rename into a separate PR ([#329](https://github.com/pulumi/pulumi-go-provider/pull/329)) to keep this PR focused on the request/response struct refactor. This should make review easier and help isolate the breaking changes.

infer/function.go
rquitales: Updated.

EronWright (approved):
LGTM

Issue #243, opened 2024-06-13
Error on provider authors when they use pulumi's Asset type 


thomas11:
As discussed in #237, the core Asset doesn't roundtrip correctly, so we introduced `types.AssetOrArchive` in this provider.

Nothing prevents a provider author from still using Asset, though. We should at least print a warning at schemagen time when they do that.

I attempted this in a simple way in [Warn about using the core Asset and Archive types](https://github.com/pulumi/pulumi-go-provider/pull/242/commits/f50c8804a534b5dc838fcf540be3872dbeecefef) but that yielded false positive warnings. The `local.Command` has Assets as an output property which is ok but triggered the warning.

iwahbe:
Right now, a provider author can have a field of the type [`*assets.Asset`](https://github.com/pulumi/pulumi/blob/80f7f05f09308a5f33dfe975ce41a6dee9b3fe0f/sdk/go/common/resource/asset/asset.go#L34-L36). When that type is translated into the Pulumi schema, it is translated into [`"pulumi.json#/Asset"`](https://www.pulumi.com/docs/iac/using-pulumi/extending-pulumi/schema/#type). When codegen generates SDKs, it materializes `"pulumi.json#/Asset"` into [`pulumi.AssetOrArchive`](https://github.com/pulumi/pulumi/blob/80f7f05f09308a5f33dfe975ce41a6dee9b3fe0f/pkg/codegen/pcl/binder_schema.go#L440-L445).

If a user then specifies an archive in their program, that gets sent back the original provider which expects an asset, which then fails to deserialize the archive (because an asset was sent).

PR #328, opened 2025-04-09
fix: error on schemagen when using resource.Asset or resource.Archive

rquitales:
**Description:**  
This PR introduces an explicit error during schema generation when `resource.Asset` or `resource.Archive` is used as the type for an input property. Provider authors should use `types.AssetOrArchive` instead, which correctly supports both types in input scenarios.

**Breaking Change**  
Input properties previously defined as `resource.Asset` or `resource.Archive` must now use `types.AssetOrArchive`. This ensures compatibility with schema generation and SDK generation workflows.

**What’s included:**  
- Added validation to schemagen to error on invalid input types  
- Updated integration tests to reflect and test this behavior

Closes: #243

### DIFF
--- a/infer/function.go
+++ b/infer/function.go
@@ -122,7 +122,7 @@ func (*derivedInvokeController[F, I, O]) GetSchema(reg schema.RegisterDerivative
 
 func objectSchema(t reflect.Type) (*pschema.ObjectTypeSpec, error) {
 	descriptions := getAnnotated(t)
-	props, required, err := propertyListFromType(t, false)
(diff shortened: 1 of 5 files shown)


rquitales:

- #328 ◀

EronWright:
I have a question, is it possible for the provider to distinguish between an Asset and an Archive? I can imagine trying to distinguish whether a given zip file was considered (in the program) an Asset versus an Archive, because that might bear on whether to extract the archive's contents.

PR #329, opened 2025-04-09
refactor(infer): rename `Call` to `Invoke` for infer.Function method

rquitales:
refactor(infer): rename `Call` to `Invoke` for infer.Function method

refactor(tests): update tests to use infer.Function.Invoke


rquitales:

- #324
    - #325
        - #327
            - #329 ◀

EronWright (commented):
From an abstract "function" perspective, `Call` is a perfectly good method name. I advocate for `Invoke` only because of the ambiguity in a Pulumi provider context.

Also useful would be to highlight that `infer.Fn` is a ["Provider Function"](https://www.pulumi.com/docs/iac/concepts/resources/functions/).  My reading of the doc is that "invokes" are called "Provider Functions" and "calls" are called "Resource Methods".

EronWright (approved):

Issue #192, opened 2024-02-12
Error on `pulumi:"secret"`


blampe:
Several times I've accidentally tagged a field as something like `pulumi:"optional,secret"` instead of `pulumi:"optional" provider:"secret"`. This generates SDKs well enough, but once you try populating this it will panic:

> panic: fatal: An assertion has failed: Unrecognized tagpart on field Foo.Token: secret

We should catch this an either error, or automatically apply the provider tag.

More generally, the pulumi/provider distinction feels like an abstraction is leaking. It would be most user-friendly to only have to worry about `pulumi:"..."` tags.

PR #330, opened 2025-04-09
fix: error during schemagen when marking a field as secret in pulumi tag

rquitales:
**Description:**
This PR improves validation of struct field tags by catching cases where `provider`-specific tag parts (like `secret`) are mistakenly placed in the `pulumi` tag. For example, the following struct tag:

```go
type myResource struct {
    myField `pulumi:"myField,optional,secret"`
}
```

should instead be written as:

```go
type myResource struct {
    myField `pulumi:"myField,optional" provider:"secret"`
}
```

Previously, this would compile and generate SDKs correctly but would lead to Pulumi program runtime panics like:

```
panic: fatal: An assertion has failed: Unrecognized tagpart on field Foo.Token: secret
```

**What’s included:**
- We now validate tag parts during schema generation and return a clear error if an invalid tag is used under the `pulumi` namespace.
- This helps provider authors catch mistakes early and avoids confusing panics for provider users later.

Closes: #192

### DIFF
--- a/internal/introspect/introspect.go
+++ b/internal/introspect/introspect.go
@@ -166,6 +166,15 @@ func ParseTag(field reflect.StructField) (FieldTag, error) {
 		}
 	}
 
+	// Determine if the provider author had accidentally marked the field as secret
+	// in the `pulumi` namespace. We need to error out if this is the case, as it will
+	// lead to a Pulumi program runtime panic.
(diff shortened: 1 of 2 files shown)


rquitales:

- #324
    - #325
        - #327
            - #328
                - #330 ◀

iwahbe (approved):

PR #335, opened 2025-04-10
[tests] bind an component input property to the child

EronWright:
Relates to https://github.com/pulumi/pulumi-go-provider/issues/337

This PR intends to test whether a component may use its inputs to set input values on its children.  In this case, we use the `MyInput` argument of `testComponent` as the prefix of the encapsulated `RandomPet`.

When one binds the `Prefix` of `RandomPet` to `MyInput`, the system appears to lock up unless the args are split out into a separate struct. For this reason, we introduce `testComponentArgs`.

To improve the usability of call_consumer as a standalone program, we add some runtime options and make some handy exports.

### DIFF
--- a/tests/grpc/call_consumer/Pulumi.yaml
+++ b/tests/grpc/call_consumer/Pulumi.yaml
@@ -1,5 +1,8 @@
 name: "call_consumer"
-runtime: python
+runtime:
+  name: python
+  options:
+    virtualenv: venv
 plugins:
   providers:
     - name: test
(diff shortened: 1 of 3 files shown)


iwahbe:
Thanks for figuring this out. Can you open an issue to track?

EronWright:
Filed: https://github.com/pulumi/pulumi-go-provider/issues/337
I'll repurpose this PR for #337.

EronWright:
Update: I found that splitting the inputs into an Args type, as is typical of infer-based components, was effective at avoiding the deadlock.

EronWright:
@iwahbe you're right, and I've now decoupled them.  This PR is more like a workaround, in that it updates the example such that it uses a true Args class for its inputs.

iwahbe (approved):
I'm happy to fix the example as long as we keep #337 open.

PR #341, opened 2025-04-16
Rewrite Call handling to use Host interface

### LINKED ISSUE #340: Rewrite `Call` handling to use Host interface
The `Provider.Call` method should not take a `pulumi.Context` in the `CallRequest`, instead the context should be manufactured later at the dispatch or infer layer. The provider should use the `Host` interface to transition into a program context before invoking the user's `Call` function.

This will ensure that the Go context can be mutated by the middleware, such as `middleware.cancel`. I would guess that cancelation doesn't work with Call at the moment.

Observe [here](https://github.com/pulumi/pulumi-go-provider/blob/f8d994264e8d2e746eb01ad57725258ee94bc026/provider.go#L823-L825) how `comProvider.Call` is being invoked as a "synchronization point".  I believe the ideal would be to call the user's `Call`  function from within the program context provided by `comProvider.Call`.  In other words, there's actually two `pulumi.Context` instances being created and we can avoid that.

EronWright:
_A follow-up to https://github.com/pulumi/pulumi-go-provider/pull/338._

This PR rewrites Call handling to be consistent with how Construct works by leveraging the Host interface.  Now we avoid the duplicative creation of two `pulumi.Context` instances, and the awkward "synchronization point".  This is accomplished by invoking the user's `Call` code naturally from within the Go SDK's `call` function.

This PR also updates the test provider's Call implementation to more fully utilize the Go SDK, e.g. it now uses Input/Output types and calls `NewCallResult`.

It integrates provider cancelation into Call, so that all outstanding Calls are canceled.  This is made possible by creating the `pulumi.Context` less eagerly than before, so that it wraps the middleware's cancelable context.

It adds support for Call to `middleware.rpc`, to be able to wrap a legacy provider that implements Call.

Finally it adds support for integration testing of Provider.Call.

_In the below diagram(s), the yellow region is where the system is in a Pulumi program context and using Go SDK features such as inputs and outputs._

```mermaid
sequenceDiagram
    actor User as Pulumi
    participant Server as Provider Server
    participant SDK as Pulumi SDK
    participant Provider

    User->>Server: Call
    activate Server
    Server->>+Provider: Provider.Call
    Provider->>Provider: getHost(ctx)
    Provider->>Server: Host.Call
    activate Server
    Server->>Server: getEngineConn
    Server->>SDK: comProvider.Call
    activate SDK
    SDK->>SDK: pulumi.NewContext
    rect rgb(255,255,191)
    SDK->>Provider: CallFunc
    activate Provider
    Provider-->>SDK: RegisterResource
    SDK-->>User: RegisterResource
    Provider->>SDK: NewCallResult
    SDK->>Provider: 
    Provider->>SDK: (result)
    deactivate Provider
    end
    SDK->>Server: (result)
    deactivate SDK
    Server->>Provider: (result)
    deactivate Server
    Provider->>Server: (result)
    deactivate Provider
    Server->>User: (response)
    deactivate Server
```

The `call_consumer` program shows the expected output, the concatenated string produced by the test provider:
```plain

View in Browser (Ctrl+O): https://app.pulumi.com/eron-pulumi-corp/call_consumer/dev/updates/18

     Type                          Name               Status              
 +   pulumi:pulumi:Stack           call_consumer-dev  created (1s)        
 +   ├─ test:index:Component       test               created (0.56s)     
 +   │  └─ random:index:RandomPet  pet                created (0.07s)     
 +   └─ random:index:RandomPet     call-pet           created (0.16s)     

Outputs:
    resp1: "bar:urn:pulumi:dev::call_consumer::test:index:Component::test:smashing-fawn"

```

Stacked on top of https://github.com/pulumi/pulumi-go-provider/pull/338.

Closes #340

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -858,7 +858,8 @@ type CallRequest struct {
 	Organization string
 }
 
-func newCallRequest(req *rpc.CallRequest, unmarshal rpcToProperty) (CallRequest, error) {
+func newCallRequest(req *rpc.CallRequest,
+	unmarshal func(s *structpb.Struct) (presource.PropertyMap, error)) (CallRequest, error) {
 	args, err := unmarshal(req.GetArgs())
 	if err != nil {
 		return CallRequest{}, fmt.Errorf("unable to convert args into a property map: %w", err)
--- a/provider.go
+++ b/provider.go
@@ -780,6 +780,9 @@ func (p *provider) Call(ctx context.Context, req *rpc.CallRequest) (*rpc.CallRes
 	if err != nil {
 		return nil, err
 	}
+
+	contract.Assertf(req.AcceptsOutputValues, "The caller must accept output values")
+
 	resp, err := p.client.Call(ctx, r)
 	if err != nil {
 		return nil, err
--- a/provider.go
+++ b/provider.go
@@ -38,10 +38,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/rpcutil/rpcerror"
-	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	comProvider "github.com/pulumi/pulumi/sdk/v3/go/pulumi/provider"
 	rpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
-	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	emptypb "google.golang.org/protobuf/types/known/emptypb"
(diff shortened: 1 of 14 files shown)


provider.go
```diff
@@ -854,13 +820,138 @@ func (p *provider) Call(ctx context.Context, req *rpc.CallRequest) (*rpc.CallRes
...
 // CallRequest represents a requested resource method invocation.
 //
 // It corresponds to [rpc.CallRequest] on the wire.
 type CallRequest struct {
-	Tok     tokens.ModuleMember
-	Args    presource.PropertyMap
-	Context *pulumi.Context
+	// the function token to invoke.
+	Tok tokens.ModuleMember
+	// the arguments for the function invocation.
```
lunaris, line 838: Possible nit: `s/function/method/g`?

provider.go
```diff
@@ -854,13 +820,138 @@ func (p *provider) Call(ctx context.Context, req *rpc.CallRequest) (*rpc.CallRes
...
+		DryRun:          req.GetDryRun(),
+		Parallel:        req.GetParallel(),
+		MonitorEndpoint: req.GetMonitorEndpoint(),
+		Organization:    req.GetOrganization(),
+	}
+	return r, nil
+}
+
+func (c CallRequest) rpc(marshal propertyToRPC) *rpc.CallRequest {
+
```
lunaris, line 904:
```suggestion
```

lunaris: There's a golangci-lint check we could add for this sort of thing. The full (somewhat aggressive) list we use in `pulumi/pulumi` that I think works well in case useful:

```
    - durationcheck                                                  
    - depguard                                                        
    - errcheck                               
    - exhaustive
    - gofumpt                                                   
    - goheader
    - goprintffuncname                                                   
    - gosec                                                            
    - govet                                                                   
    - importas                                                           
    - ineffassign                   
    - lll
    - misspell
    - nakedret  
    - nolintlint                      
    - paralleltest                                                       
    - perfsprint       
    - prealloc                                                                                         
    - revive
    - unconvert                                               
    - unused                                          
    - usetesting                                                    
    - wastedassign   
    - whitespace   
```

`whitespace` is the one in question here I think.

provider.go
```diff
@@ -854,13 +820,138 @@ func (p *provider) Call(ctx context.Context, req *rpc.CallRequest) (*rpc.CallRes
...
+	fromUrns := func(urns []presource.URN) []string {
+		r := make([]string, len(urns))
+		for i, urn := range urns {
+			r[i] = string(urn)
+		}
+		return r
+	}
+
+	// Marshal the args.
```
lunaris, line 913:
```suggestion
```
Not sure this is necessary? Same for similar comments about unmarshalling later, but happy either way

lunaris (approved):
Nice!

provider.go
EronWright: We need 100% fidelity going in/out of the rpc representation, to support `middleware.rpc` which calls into a legacy provider.

provider.go
EronWright: I got the text from upstream:
https://github.com/pulumi/pulumi/blob/0ff190607702431a286161d8847902e2238e39f2/proto/pulumi/provider.proto#L502-L512

PR #344, opened 2025-04-18
Migrate `resource.PropertyValue` to `property.Value` in public APIs

iwahbe:
We believe that `property.Value` is a safer and less mistake prone API compared to `resource.PropertyValue`. While we break all the things, let's take the opportunity to use the better representation of pulumi values. Critically, `property.Value` is an immutable data type. **We require an immutable value representation to handle a native set type**, which we would like to add in the future.

This PR changes all public APIs in this repo to use `property.Value` and `property.Map` instead of `resource.PropertyValue` and `resource.PropertyMap`. Internal values were migrated only as convenient, primarily to reduce the size of this already large PR. 

Fixes https://github.com/pulumi/pulumi-go-provider/issues/347

### DIFF
--- a/tests/grpc/call_test.go
+++ b/tests/grpc/call_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/config"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/pulumi/pulumi/sdk/v3/go/property"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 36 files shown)


tests/grpc/call_test.go
```diff
@@ -110,14 +111,11 @@ func TestCall(t *testing.T) {
...
 
 		return p.CallResponse{
-			Return: resource.PropertyMap{
-				"r1": resource.NewProperty(resource.Output{
-					Element: resource.NewProperty("e1"),
-					Dependencies: []resource.URN{
-						"urn1", "urn2",
-					},
+			Return: property.NewMap(map[string]property.Value{
+				"r1": property.New(property.Computed).WithDependencies([]resource.URN{
```
iwahbe, line 115: One issue I have with `resource.PropertyValue` is that it makes it very easy to construct the wrong thing. Here, the `Element: resource.NewProperty("e1")` doesn't do anything since `Known: true` was not set. This value is correctly translated to `property.New(property.Computed)`.

Frassle: 🤦‍♂️ something something structs and default zeros

EronWright (approved):
I think this is ready to go.

PR #349, opened 2025-04-30
Test flake: TestRapidDeepEqual

EronWright:
A fix for test flakiness, which seems to be happening quite regularly, e.g:
- https://github.com/pulumi/pulumi-go-provider/actions/runs/14743562356/job/41386941559?pr=344

```
           Failed test output:
        putil_test.go:44: [rapid] draw distinct: []resource.PropertyValue{resource.PropertyValue{V:resource.Output{Element:resource.PropertyValue{V:resource.Computed{Element:resource.PropertyValue{V:interface {}(nil)}}}, Known:true, Secret:false, Dependencies:[]urn.URN{"̇𜼿", "#\ufeff#{~ᾏʱA ", "*A\u00a0̆'ᾘa"}}}, resource.PropertyValue{V:resource.Computed{Element:resource.PropertyValue{V:interface {}(nil)}}}}
        putil_test.go:45: 
            	Error Trace:	/home/runner/work/pulumi-go-provider/pulumi-go-provider/internal/putil/putil_test.go:45
            	            				/home/runner/go/pkg/mod/pgregory.net/rapid@v1.1.0/engine.go:368
            	            				/home/runner/go/pkg/mod/pgregory.net/rapid@v1.1.0/engine.go:232
            	            				/home/runner/go/pkg/mod/pgregory.net/rapid@v1.1.0/engine.go:138
            	Error:      	Should be false
            	Test:       	TestRapidDeepEqual/distinct-values
```

The problem seems to be that the property value generator produces values that putil would consider equal yet have a distinct key, and so rapid expects the values to be distinct.  The values are considered equal because of _folding_, which simplifies values such as outputs nested within outputs and secrets nested within outputs.

The clear solution is to apply folding when generating the key of a given value.  Note that folding isn't applied to the value under test, it is simply done to generate a distinct key.  The `foldOutputValue` method comes from putil's internals.

An alternative would be to change the generator to not produce "foldable" (un-normalized) values, but it seemed simpler to simply improve the fidelity of the keyfunc.

### DIFF
--- a/internal/putil/putil_test.go
+++ b/internal/putil/putil_test.go
@@ -37,8 +37,8 @@ func TestRapidDeepEqual(t *testing.T) {
 	t.Run("distinct-values", rapid.MakeCheck(func(t *rapid.T) { //nolint:paralleltest
 		// keyFn is how [rapid] determines that values are distinct.
 		//
-		// We do this by calling [fmt.Stringer.String] on the normalized form of v.
-		keyFn := func(v r.PropertyValue) string { return normalize(v).String() }
+		// We do this by calling [fmt.Stringer.String] on the folded form of v.
+		keyFn := func(v r.PropertyValue) string { return foldOutputValue(v).String() }
 		values := rapid.SliceOfNDistinct(
 			rresource.PropertyValue(5), 2, 2, keyFn,
 		).Draw(t, "distinct")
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
It feels bad to include this logic into test code (since its part of what the function is testing), but I'm OK here since (1) flaky tests are super annoying and (2) `putil.DeepEquals` is stable and (3) this will all go away when `infer` switches to using `propert.Value` internally.

PR #350, opened 2025-04-30
Remove `InputDependencies` from Provider interface

### LINKED ISSUE #346: Remove InputDependencies from ConstructOptions
See discussion here: https://github.com/pulumi/pulumi-go-provider/pull/338/files/cbeef2a5d61e3b5320bfa0274508a672c83d9530#r2048882567

EronWright:
This PR removes the formal "dependencies" map that often accompanies property maps on Provider requests and responses. Any dependences on an incoming RPC request are simply folded into their respective properties. The resultant RPC response never contains a dependencies map, and the provider asserts that the caller MUST accept output values.

The request/response structs are changing like this:
image

The middleware.rpc package takes pains to downgrade input dependencies based on the wrapped provider's capabilities (Call and Construct ONLY). The wrapped provider MAY return a dependencies map in all modes.

**To discuss**: we know that dependency maps are often generated by walking an output value (recursively). When we use upgrade logic to re-attach the dependencies to the property values, does that cause some "garbage" to build up on those values?  Should the caller provide a dependency map only in a downgrade situation?  When may it be ignored? Should we filter out the deps that are present on the children?

Here, for example, we re-attach the dependencies map to the property values.  If those deps were extracted from deep children, now they're attached directly to the top-level value. Very concerning to me.
https://github.com/pulumi/pulumi-go-provider/pull/350/files#diff-1bb2cdf24b7aa1b1986f79d4ea823ec921a95ba94f6436119cb2a74c21ebff82R876-R883

on-wire:
```plain
v
├── k1 (urn1)
└── k2
    └── k3 (urn2)
```

after upgrade:
```plain
v (urn1, urn2)
├── k1 (urn1)
└── k2
    └── k3 (urn2)
```

My best idea here is to re-attach ONLY the deps that aren't present on the children. WDYT?

Update: It is assumed that, for a given value, the dependencies are either within the value itself OR in the dependency map. So, if the value has any dependencies, you needn't incorporate the map, and doing so would have the erroneous effect of folding the child dependencies into the parent.

Closes #346

### DIFF
--- a/internal/putil/putil.go
+++ b/internal/putil/putil.go
@@ -17,9 +17,11 @@ package putil
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/urn"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/property"
 )
@@ -241,6 +243,29 @@ func ToUrns(s []string) []resource.URN {
 	}
 	return r
 }
+func ToUrns2(s []string) []urn.URN {
+	r := make([]urn.URN, len(s))
+	for i, a := range s {
+		r[i] = urn.URN(a)
+	}
+	return r
+}
+
+func FromUrns(urns []resource.URN) []string {
+	r := make([]string, len(urns))
+	for i, urn := range urns {
+		r[i] = string(urn)
+	}
+	return r
+}
+
+func FromUrns2(urns []urn.URN) []string {
+	r := make([]string, len(urns))
+	for i, urn := range urns {
+		r[i] = string(urn)
+	}
+	return r
+}
 
 // Walk traverses a property value along all paths, performing a depth first search.
 func Walk(v property.Value, f func(property.Value) (continueWalking bool)) bool {
@@ -263,4 +288,43 @@ func Walk(v property.Value, f func(property.Value) (continueWalking bool)) bool 
 		}
 	}
 	return true
+}
+
+// GetPropertyDependencies gathers (deeply) the dependencies of the given property value.
+func GetPropertyDependencies(v property.Value) []urn.URN {
+	var deps []urn.URN
+	Walk(v, func(v property.Value) (continueWalking bool) {
+		deps = append(deps, v.Dependencies()...)
+		return true
+	})
+	slices.Sort(deps)
+	return slices.Compact(deps)
+}
+
+// MergePropertyDependencies merges the given dependencies into the property map.
+//
+// Apply the dependencies to the property map, only for "legacy" values that don't have output dependencies.
+// The caller may fold the output dependencies of a value's children into the dependencies map,
+// and we seek to avoid treating those as true dependencies of the top-level value.
+func MergePropertyDependencies(m property.Map, dependencies map[string][]urn.URN) property.Map {
+	if len(dependencies) == 0 {
+		return m
+	}
+	_m := m.AsMap()
+	for name, deps := range dependencies {
+		if v, ok := _m[name]; ok {
+			vdeps := GetPropertyDependencies(v)
+			deps := slices.DeleteFunc(slices.Clone(deps), func(dep urn.URN) bool {
+				return slices.Contains(vdeps, dep)
+			})
+			deps = mergePropertyDependencies(append(v.Dependencies(), deps...))
+			_m[name] = v.WithDependencies(deps)
+		}
+	}
+	return property.NewMap(_m)
+}
+
+func mergePropertyDependencies(deps []urn.URN) []urn.URN {
+	slices.Sort(deps)
+	return slices.Compact(deps)
 }
--- a/internal/putil/putil.go
+++ b/internal/putil/putil.go
@@ -231,3 +231,12 @@ func ParseProviderReference(s string) (resource.URN, resource.ID, error) {
 func FormatProviderReference(urn resource.URN, id resource.ID) string {
 	return fmt.Sprintf("%s%s%s", urn, resource.URNNameDelimiter, id)
 }
+
+// ToUrns converts a slice of strings to a slice of URNs.
+func ToUrns(s []string) []resource.URN {
+	r := make([]resource.URN, len(s))
+	for i, a := range s {
+		r[i] = resource.URN(a)
+	}
+	return r
+}
--- a/internal/putil/putil.go
+++ b/internal/putil/putil.go
@@ -236,14 +236,7 @@ func FormatProviderReference(urn resource.URN, id resource.ID) string {
 }
 
 // ToUrns converts a slice of strings to a slice of URNs.
-func ToUrns(s []string) []resource.URN {
-	r := make([]resource.URN, len(s))
-	for i, a := range s {
-		r[i] = resource.URN(a)
-	}
-	return r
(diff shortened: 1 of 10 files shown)


iwahbe:
Similar to https://github.com/pulumi/pulumi-go-provider/issues/346, if we can infer what we need from the dependencies found in the return value, we should just remove `ReturnDependencies` from `CallResponse`.

tests/grpc/call_test.go
```diff
@@ -115,7 +116,11 @@ func TestCall(t *testing.T) {
 				"r1": property.New(property.Computed).WithDependencies([]resource.URN{
 					"urn1", "urn2",
 				}),
+				"r2": property.New("s"),
 			}),
+			ReturnDependencies: map[string][]resource.URN{
+				"r2": {resource.URN("urn1"), resource.URN("urn2")},
+			},
```
EronWright, line 123: Here we add another return property `r2` that is encoded in an old style, e.g. as may be returned from a wrapped legacy provider. Of course, `r1` is in the new style.

EronWright:
As you know, my rationale is to keep it DRY, to minimize complexity, and to ensure high fidelity for wrapped legacy providers.   

I went back and updated call_test.go to produce a more realistic mix of old- and new-style return values, and found that the upgrade logic was actually introducing a bug (see inline). Does that help make the case here?

EronWright:
PR was totally reworked as per discussion, please take a look.

internal/putil/putil_test.go
```diff
@@ -171,3 +172,35 @@ func TestParseProviderReference(t *testing.T) {
...
+				"k": property.New("s").WithDependencies([]r.URN{"s"}),
+			}).WithDependencies([]r.URN{"k"}),
+			want: []r.URN{"k", "s"},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.v.GoString(), func(t *testing.T) {
+			t.Parallel()
+			var got []r.URN
+			putil.Walk(tc.v, func(v property.Value) (continueWalking bool) {
```
iwahbe, line 199: If `putil.Walk` is going to return something, shouldn't we test what it returns?

middleware/rpc/provider.go
```diff
@@ -361,6 +414,24 @@ func checkFailures(resp []*rpc.CheckFailure) []p.CheckFailure {
...
+func getPropertyDependencies(v property.Value) []string {
+	var deps []string
+	putil.Walk(v, func(v property.Value) (continueWalking bool) {
+		for _, v := range v.Dependencies() {
+			deps = append(deps, string(v))
+		}
+		return true
+	})
+	slices.Sort(deps)
+	return slices.Compact(deps)
```
iwahbe, line 427:
```suggestion
	return mergePropertyDependencies(deps)
```

middleware/rpc/provider.go
```diff
@@ -288,6 +330,17 @@ func Provider(server rpc.ResourceProviderServer) p.Provider {
 				return p.CallResponse{}, err
 			}
 
+			// upgrade the return dependencies if the provider doesn't support output values,
+			// in which case [rpcResp.ReturnDependencies] has meaningful information.
+			_return := resp.Return.AsMap()
+			for name, deps := range rpcResp.GetReturnDependencies() {
+				if v, ok := _return[name]; ok {
+					deps := mergePropertyDependencies(append(v.Dependencies(), putil.ToUrns(deps.GetUrns())...))
```
iwahbe, line 338: Good catch! Property dependencies should just do this: https://github.com/pulumi/pulumi/pull/19423.

provider.go
```diff
@@ -1570,9 +1453,20 @@ func (p *provider) Construct(ctx context.Context, req *rpc.ConstructRequest) (*r
...
+
+	state, err := p.asStruct(resp.State)
+	if err != nil {
+		return nil, err
+	}
 
-	return result.rpc(p.asStruct), err
+	return &rpc.ConstructResponse{
+		Urn:   string(resp.Urn),
+		State: state,
```
iwahbe, line 1468: If `req.AcceptsOutputValues` is false, then we need to reconstruct `StateDependencies` here.

provider.go
```diff
@@ -812,6 +795,10 @@ func (h *host) Call(ctx context.Context, req CallRequest, call comProvider.CallF
 	if err != nil {
 		return CallResponse{}, err
 	}
+
+	// note that req.ReturnDependencies is silently discarded
+	// because the information is simply derived from the property values in r.Return.
+
```
iwahbe, line 801: If `req.AcceptsOutputValues` is false, we will need to reconstruct `ReturnDependencies` when sending out.

middleware/rpc/provider.go
EronWright: Unfortunately the types don't match:
```
cannot use mergePropertyDependencies(deps) (value of type []resource.URN) as []string value in return statementcompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)
```

iwahbe: Unfortunate.

provider.go
EronWright: That's true in general but not relevant here because in host.Call we're unmarshaling from rpc, not marshaling to rpc.  Later, in provider.Call, the rpc result is constructed.

provider.go
EronWright: This implementation does not accept requests where AcceptsOutputValues is false.
```go
contract.Assertf(req.AcceptsOutputValues, "The caller must accept output values")

```

iwahbe: As long as we fail loudly, I'm fine with this as is.

iwahbe (approved):

internal/putil/putil.go
```diff
@@ -264,3 +289,42 @@ func Walk(v property.Value, f func(property.Value) (continueWalking bool)) bool
...
+	})
+	slices.Sort(deps)
+	return slices.Compact(deps)
+}
+
+// MergePropertyDependencies merges the given dependencies into the property map.
+//
+// Apply the dependencies to the property map, only for "legacy" values that don't have output dependencies.
+// The caller may fold the output dependencies of a value's children into the dependencies map,
+// and we seek to avoid treating those as true dependencies of the top-level value.
```
iwahbe, line 293: I don't know what this means.

Are you saying, given a map `m` and a dependency map `deps`: 
```
m = { Dependencies: ["urn1"], Values: { "n": { Dependencies: ["urn2"] } } }
deps = { "m": ["urn3"] }
```

that we want to see 

```
MergePropertyDependencies(m, deps) = 
	{ Dependencies: ["urn1", "urn3"], Values: { "n": { Dependencies: ["urn2"] } } }
```

An example here would make it much easier to understand.

internal/putil/putil.go
```diff
@@ -241,6 +243,29 @@ func ToUrns(s []string) []resource.URN {
...
 }
+func ToUrns2(s []string) []urn.URN {
+	r := make([]urn.URN, len(s))
+	for i, a := range s {
+		r[i] = urn.URN(a)
+	}
+	return r
+}
+
+func FromUrns(urns []resource.URN) []string {
```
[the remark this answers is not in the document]
EronWright: I cleaned it up but the overall transition from resource.URN to urn.URN will need to be a future project.

internal/putil/putil.go
EronWright: Yes. See the test cases in putil_test.go for some examples.

justinvp (commented):
Looks reasonable to me

PR #352, opened 2025-04-30
Rename Olds/News -> State/Inputs

blampe:
Olds/News is a Pulumi naming convention that isn't immediately obvious to newcomers.

We're already using State/Inputs terminology, which is much more obvious. Let's rename these for consistency.

Edit: We could also go the other way and rename State/Inputs to → Olds/News if we want to keep things as close as possible to the protocol.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -63,8 +63,8 @@ type GetSchemaResponse struct {
 
 type CheckRequest struct {
 	Urn        presource.URN
-	Olds       property.Map
-	News       property.Map
+	State      property.Map
+	Inputs     property.Map
 	RandomSeed []byte
 }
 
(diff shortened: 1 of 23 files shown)


iwahbe (commented):
If we're going to do this, we should do it for `Provider`, not just within `infer`.

EronWright (approved):

iwahbe (approved):
CC @thoward

provider.go
```diff
@@ -262,8 +261,8 @@ type ReadResponse struct {
 type UpdateRequest struct {
 	ID            string        // the ID of the resource to update.
 	Urn           presource.URN // the Pulumi URN for this resource.
-	Olds          property.Map  // the old values of provider inputs for the resource to update.
-	News          property.Map  // the new values of provider inputs for the resource to update.
+	State         property.Map  // the old values of provider inputs for the resource to update.
```
EronWright, line 264: "old state"

EronWright (approved):

EronWright:
Consider renaming `Preview` to `DryRun`.

PR #355, opened 2025-05-01
Remove resourcex

### LINKED ISSUE #353: Move `resourcex` to its own repo
## Hello!

## Issue details

`resourcex` contains experimental `resource.PropertyValue` utilities. `pulumi-go-provider` is a poor home for it, since `pulumi-go-provider` no longer exposes `resource.PropertyValue`. As part of v1.0.0, we should take the opportunity to clean up here.

We can move it to it's own repo, or we can fold it into a more appropriate home.

blampe:
I don't see any consumers of resourcex which are also themselves go-providers:

https://github.com/search?q=%22github.com%2Fpulumi%2Fpulumi-go-provider%2Fresourcex%22&type=code

Those consumers can continue to use resourcex from the existing v0.26.0 release.

If there's a need we can revive this and put it somewhere else. 

Fixes #353


iwahbe (approved):
🔥🔥🔥

provider.go
```diff
@@ -787,15 +786,11 @@ func (p *provider) Call(ctx context.Context, req *rpc.CallRequest) (*rpc.CallRes
...
-			}
-			for _, dep := range v.OutputValue().Dependencies {
-				urns = append(urns, string(dep))
-			}
-		})
+		deps := v.Dependencies()
+		urns := make([]string, 0, len(deps))
+		for _, u := range v.Dependencies() {
+			urns = append(urns, string(u))
+		}
```
EronWright, line 793: (squinting)
did the deps become shallow?

PR #356, opened 2025-05-01
Error on Asset types

### LINKED ISSUE #243: Error on provider authors when they use pulumi's Asset type

blampe:
Revives https://github.com/pulumi/pulumi-go-provider/pull/328

Fixes #243


infer/schema.go
```diff
@@ -126,13 +126,25 @@ func getResourceSchema[R, I, O any](isComponent bool) (schema.ResourceSpec, mult
...
+	// We will returrn an error if resource.Asset or resource.Archive is used directly for an input.
+	// pulumi/pulumi-go-provider#243
+	if propType == inputType && t == reflect.TypeOf(resource.Asset{}) {
+		return schema.TypeSpec{},
+			fmt.Errorf("resource.Asset is not a valid input type, please use types.AssetOrArchive instead")
+	}
+	if propType == inputType && t == reflect.TypeOf(resource.Archive{}) {
+		return schema.TypeSpec{},
+			fmt.Errorf("resource.Archive is not a valid input type, please use types.AssetOrArchive instead")
+	}
```
iwahbe, line 144: AFAIK archives are correctly enforced, and are thus valid input types.

Under Pulumi's current type system, an asset is an archive but an archive is not an asset.

iwahbe (approved):

PR #363, opened 2025-05-06
Write pulumi-go-provider repo’s integration README

EronWright:
Closes https://github.com/pulumi/home/issues/3960

PR #364, opened 2025-05-06
Split up BuildAndRun

blampe:
It's very common (especially in testing) to construct a provider and then pass it around before acting on it.

https://github.com/pulumi/pulumi-docker-build/blob/30a01b689388605b53cf10594ad593c9f270cf15/provider/internal/provider.go#L68

https://github.com/pulumi/pulumi-provider-boilerplate/blob/5c43e6c38e62b35b3352dc535f6e83c6cf4ab32f/provider/provider.go#L28

This means `BuildAndRun` is too restrictive, so this PR breaks out `Build` into a standalone method which returns an error if validation fails.

`Run` is added to go-provider's `Provider` struct to spare the user the unintuitive `provider.Main` call. Name and Version make this a little kludgy because we don't currently track those on the `Provider` itself.

### DIFF
--- a/infer/provider_builder.go
+++ b/infer/provider_builder.go
@@ -38,29 +38,30 @@ type ProviderBuilder struct {
 //
 // This is an example of how to create a simple provider with a single component resource:
 //
-//		type RandomComponent struct {
-//			pulumi.ResourceState
-//			RandomComponentArgs
-//		 	Password        pulumi.StringOutput `pulumi:"password"`
-//		}
+//	type RandomComponent struct {
+//	    pulumi.ResourceState
+//	    RandomComponentArgs
 //
-//		type RandomComponentArgs struct {
-//			Length pulumi.IntInput `pulumi:"length"`
-//		}
+//	    Password pulumi.StringOutput `pulumi:"password"`
+//	}
 //
-//		func NewMyComponent(ctx *pulumi.Context, name string,
-//				compArgs RandomComponentArgs, opts ...pulumi.ResourceOption) (*RandomComponent, error) {
-//			// Define your component constructor logic here.
-//		}
+//	type RandomComponentArgs struct {
+//	    Length pulumi.IntInput `pulumi:"length"`
+//	}
 //
-//		func main() {
-//			p, _ := infer.NewProviderBuilder().
-//				WithComponents(
-//					infer.Component(NewMyComponent),
-//				).
-//				Build()
-//	        p.Run(context.Background(), "go-components", "v0.0.1")
-//		}
+//	func NewMyComponent(ctx *pulumi.Context, name string,
+//	    compArgs RandomComponentArgs, opts ...pulumi.ResourceOption) (*RandomComponent, error) {
+//	    // Define your component constructor logic here.
+//	}
+//
+//	func main() {
+//	    p, _ := infer.NewProviderBuilder().
+//	        WithComponents(
+//	            infer.Component(NewMyComponent),
+//	        ).
+//	        Build()
+//	    p.Run(context.Background(), "go-components", "v0.0.1")
+//	}
 //
 // Please note that the initial defaults provided by this function may change with future releases of
 // this framework. Currently, we are setting the following defaults:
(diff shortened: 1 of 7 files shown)


infer/provider_builder_test.go
```diff
@@ -330,27 +335,32 @@ func TestValidate(t *testing.T) {
...
 	}(errChan)
 
-	select {
-	case err := <-errChan:
-		require.NoError(t, err, "provider startup should not fail")
-	case <-time.After(5 * time.Second):
-		return // The provider started successfully, so we can return.
-	}
+	err = <-errChan
+	assert.NoError(t, err, "provider startup should not fail")
```
blampe, line 365: This is using https://github.com/pulumi/pulumi/pull/19445.

The assertion is stronger because we're now guaranteed the provider executed and return a nil error.

infer/provider_builder_test.go
```diff
@@ -330,27 +335,21 @@ func TestValidate(t *testing.T) {
...
+		WithVersion("1.0.0").
+		Build()
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	err = p.Run(ctx)
+
+	assert.NoError(t, err, "provider startup should not fail")
```
blampe, line 354: This is using https://github.com/pulumi/pulumi/pull/19445.

The assertion is stronger because we're now guaranteed the provider executed and returned a nil error.

provider.go
```diff
@@ -373,6 +373,9 @@ type Provider struct {
 
 	// Components Resources
 	Construct func(context.Context, ConstructRequest) (ConstructResponse, error)
+
+	Name    string
+	Version string
```
iwahbe, line 378: This change needs a bit more thought in my opinion. We can think about that in a separate PR by changing the signature of `(Provider).Run` to `func(ctx context.Context, name, version string) error` for now.

iwahbe (commented):
I'm supportive of the change described in the PR description. This PR does a lot more then that though. Can we cut this PR back to splitting up BuildAndRun?

provider.go
blampe: That's fine I guess, I just really want to avoid a final API that looks like this:

```go
p, _ := builder.WithName(name).WithVersion(version).Build()
p.Run(ctx, name, version)
```

An alternative would be to drop `WithName` and `WithVersion` but I probably prefer to keep them.

AFAICT name is only used in pu/pu for tracing, and go-provider is adding them to the context's RuntimeInfo.

EronWright:
```go
// RunProviderF runs a provider with the given name and version.
func RunProviderF(name, version string, providerF func(*pprovider.HostClient) (Provider, error)) error {
	return pprovider.Main(name, func(host *pprovider.HostClient) (rpc.ResourceProviderServer, error) {
		provider, err := providerF(host)
		if err != nil {
			return nil, err
		}
		return newProvider(name, version, provider.WithDefaults())(host)
	})
}
```

And a pure `Build()`:
https://github.com/pulumi/pulumi-go-provider/pull/363/commits/39d6bf6c733d57f99ed59dbdcae2bb1c0eb4dea6

EronWright:
To make this complete for integration testing, we would need to change the signature of NewServer, e.g 
```
type ProviderF func(*pprovider.HostClient) Provider

func NewServerWithOptions(ctx context.Context, pkg string, version semver.Version, providerF ProviderF,
	opts ...ServerOption,
) Server {
        ...
	host := newHost(ctx, o.mocks)
        provider := providerF(host)
	return &server{p.RunInfo{
		PackageName: pkg,
		Version:     version.String(),
	}, host, provider.WithDefaults(), ctx}
}
```

infer/provider_builder.go
```diff
@@ -40,30 +38,29 @@ type ProviderBuilder struct {
...
-//			).
-//			BuildAndRun()
-//	}
+//		func main() {
+//			p, _ := infer.NewProviderBuilder().
+//				WithComponents(
+//					infer.Component(NewMyComponent),
+//				).
+//				Build()
+//	        p.Run(context.Background(), "go-components", "v0.0.1")
```
EronWright, line 62: The formatting seems wrong here.

EronWright (approved):
Great stuff!

PR #365, opened 2025-05-06
Infer from values instead of types

blampe:
This changes up our API to infer the provider's schema from, and to dispatch requests to, concrete instances instead of abstract types. For example, hooking up a resource now looks like `infer.Resource(&R{})` instead of `infer.Resource[*R]()`. It's a small syntactical difference but it gives the framework a lot more power and flexibility.

To round out this capability for both custom resources and component resources, this PR re-introduces the option of using a struct to define a component resource. This option was recently removed in https://github.com/pulumi/pulumi-go-provider/pull/313 in favor of a functional alternative that resembles that of a "local" component. With this PR, both styles are supported.

The first benefit is call signatures will feel more familiar to most Go developers. The reflect-ish `f[T]()` style makes types a very front-and-center concern while working, but our users just want to plug into some CRUD methods. We want something that gets out of the user's way, and simply passing something that implements an interface (`f(r)`) feels much more natural.

The second benefit is the user is now able to attach some "state" to their controllers (e.g. `&R{log: logger}`), which they can leverage during `Create`, `Update`, etc. This means it's now possible to unit test these providers with mock clients injected, whereas previously the only option available was integration testing. More specifically, users can now test provider logic using their schema's own Go structs without needing to dip into property maps. [Here](https://github.com/pulumi/pulumi-docker-build/pull/543/files#diff-39cd6456a9a1e4fd2f97f74f65ef2eeb5947907ba7e221121fba60c4e5c0c756L393-L413) is an example of an integration test modified to be a unit test instead.

As a corollary of invoking methods on the provided instances, it's also now easier to mock resources/components/functions. (This PR includes an [example](https://github.com/pulumi/pulumi-docker-build/pull/543/files#diff-39cd6456a9a1e4fd2f97f74f65ef2eeb5947907ba7e221121fba60c4e5c0c756R356-R366) where I use some mocks to confirm all of our methods are invoked on the user's instance.)

The last benefit that comes to mind is more for maintainers. Where previously we always had to go from { type → reflect → interface check }, there's now less of a need for the intermediate reflection layer. In other words, { instance → interface check } is now possible. Reducing the amount of reflection in the library is a big win for maintainability IMO.


iwahbe:
Looking at the full definition of `infer.Component`:
https://github.com/pulumi/pulumi-go-provider/blob/63fe29dd458d674a39ad890f04b2c56bb18a501d/infer/component.go#L56-L58

We see that the specific value of it's input is captured. The function (a Go value) that is passed in is what will end up getting called. This is not the same for `infer.Resource`:

https://github.com/pulumi/pulumi-go-provider/blob/63fe29dd458d674a39ad890f04b2c56bb18a501d/infer/resource.go#L936-L938

We rely only on the type here, which is what the signature of `infer.Resource` expresses. Given what values `infer.Resource` conceptually accepts, I think the existing signature is actually more idiomatic.

Consider the recently added `reflect.TypeFor` function:

```go
func TypeFor[T any]() Type 
```

Because `TypeFor` actually takes a type as it's argument, it is expressed as a type parameter. This was added after `reflect.TypeOf` (`func TypeOf(i any) Type`) was already in the standard library.

blampe:
> We see that the specific value of it's input is captured. The function (a Go value) that is passed in is what will end up getting called. This is not the same for `infer.Resource`

Yes, another way to frame this is that we should be capturing the resource/config/function.

Right now we don't use constructors for these things, so we don't have a straightforward way to attach some state to the method receivers you get. For example in tests I want my `*p` receiver to have a mock `p.client` but in prod I want it to actually exercise APIs. 

The [workaround](https://github.com/pulumi/pulumi-docker-build/blob/30a01b689388605b53cf10594ad593c9f270cf15/provider/internal/provider_test.go#L81-L85) I came up with in the d-b provider was to cram a mock into some context middleware and then have every handler check to see if a mock was present. Yuck!

It's a lot more user-friendly to accept something from the user which we then allow them to use later. That plumbing isn't in this PR, but it becomes possible.

EronWright (commented):
This does seem like a good improvement that would allow an infer provider to share internal state with the resource implementation code.

EronWright:
My inclination is to punt this because it is quite expensive to change at this point, and there seems to be workarounds like using Go context. WDYT?

blampe:
> My inclination is to punt this because it is quite expensive to change at this point, and there seems to be workarounds like using Go context. WDYT?

I agree we don't need it for an RC but I would like to still consider it for 1.0. If the concern is docs I can volunteer to clean all those up.

iwahbe:
> I agree we don't need it for an RC but I would like to still consider it for 1.0. If the concern is docs I can volunteer to clean all those up.

IMO a rc should be, if no problems are found, releasable.

iwahbe:
My concern now is that `infer.Resource(MyResource{mock: someMock})` doesn't work, and people will expect it to. If we wanted to take this change, then I think we should do so with the underlying functional change to support it.

EronWright:
Am curious, would components also have a useful receiver for parity with resources?

iwahbe:
> Am curious, would components also have a useful receiver for parity with resources?

No, but you can fake it with lambdas. It's a lot harder to fake with struct methods.

blampe:
> My concern now is that infer.Resource(MyResource{mock: someMock}) doesn't work, and people will expect it to. If we wanted to take this change, then I think we should do so with the underlying functional change to support it.

That's reasonable. I've updated things to the point where tests are passing, and I'll try plugging this into docker-build to confirm it helps simplify things. Then I can circle back here and add some more tests.

I remember now one of the downsides of the context middleware workaround is that it forces you to use the integration helper, because the request needs to visit all the middleware layers. That means inputs are all in terms of property maps. This is definitely _less_ painful with the new `property.Value`, but there's something very appealing about being able to unit test a provider's common-case behavior with plain Go structs, and integration reserved for edge cases.

> Am curious, would components also have a useful receiver for parity with resources?

We could make them more similar to invokes, which do have a useful receiver (the struct implementing `Invoke`). 

Right now `Component` takes a closure and basically derives a corresponding struct. We could instead have the user pass a struct in the same way as invokes, and `Component` could still wrap its `Construct` method.

iwahbe:
> We could make them more similar to invokes, which do have a useful receiver (the struct implementing `Invoke`).
> 
> Right now `Component` takes a closure and basically derives a corresponding struct. We could instead have the user pass a struct in the same way as invokes, and `Component` could still wrap its `Construct` method.

We just moved away from that pattern. You can get local state by passing a lambda to `infer.Component`:

```go
func componentWithState(state state) ComponentFn[A, R] {
	return func(
		ctx *pulumi.Context, name string, args A, opts ...pulumi.ResourceOption,
	) (R, error) {
		return myComponentFunc(ctx, name, args, state, opts)
	}
}
```

blampe:
> We just moved away from that pattern. You can get local state by passing a lambda to infer.Component:

What was the reasoning behind that? Personally I dislike closures because they can't be inspected by a debugger, and testing their internal state is a lot more complicated compared to checking some fields on a struct.

In any case it seems odd to have invokes follow a different pattern, no?

FWIW I put together https://github.com/pulumi/pulumi-docker-build/pull/543 to demonstrate what unit testing looks like with this API change. It helped me find an issue with one of the existing tests, and made me realize Read needs better documentation.

infer/tests/provider_test.go
```diff
@@ -581,23 +581,23 @@ func providerOpts(config infer.InferredConfig) infer.Options {
...
+			infer.Resource(&WithDefaults{}),
+			infer.Resource(&ReadEnv{}),
+			infer.Resource(&Recursive{}),
+			infer.Resource(&ReadConfig{}),
+			infer.Resource(&ReadConfigCustom{}),
+			infer.Resource(&CustomCheckNoDefaults{}),
 		},
 		Components: []infer.InferredComponent{
 			infer.Component(NewRandomComponent),
 			infer.Component(NewReadConfigComponent),
```
EronWright, line 597: I would consider it a blocker if we don't solve for component creation.  Hopefully the functional style can be a convenience that is also supported.

Some suggestions:
```go
Components: []infer.InferredComponent{
  infer.ComponentF(NewRandomComponent),
  infer.Component(&RandomComponent{}),
  infer.ComponentResource(&RandomComponent{}),
}
```

EronWright (changes requested):

infer/tests/provider_test.go
blampe: I was thinking the same, I like `ComponentF`.

infer/function.go
```diff
@@ -154,12 +167,7 @@ func (r *derivedInvokeController[F, I, O]) Invoke(ctx context.Context, req p.Inv
 		return p.InvokeResponse{}, fmt.Errorf("unable to apply defaults: %w", err)
 	}
 
-	var f F
-	// If F is a *struct, we need to rehydrate the underlying struct
-	if v := reflect.ValueOf(f); v.Kind() == reflect.Pointer && v.IsNil() {
-		f = reflect.New(v.Type().Elem()).Interface().(F)
-	}
-	o, err := f.Invoke(ctx, FunctionRequest[I]{Input: i})
+	o, err := r.receiver.Invoke(ctx, FunctionRequest[I]{Input: i})
```
blampe, line 170: You can see how this also starts to help make it easier to reason about the framework.

infer/tests/types_test.go
```diff
@@ -76,3 +78,135 @@ func TestOmittingAssetTypes(t *testing.T) {
 	require.Error(t, err)
 	require.Contains(t, err.Error(), "is not a valid input type, please use types.AssetOrArchive instead")
 }
+
+type (
+	TestArgs  struct{}
+	TestState struct{}
+)
+
+func TestReceivers(t *testing.T) {
```
blampe, line 87: An example of unit testing with some mocks.

tests/config_test.go
```diff
@@ -201,21 +201,22 @@ func (c *config) Check(
...
 
-	s, err := integration.NewServer(t.Context(),
-		"test",
-		semver.MustParse("0.0.0"),
-		integration.WithProvider(infer.Provider(infer.Options{
-			Config: infer.Config[*config](),
-		})),
-	)
-	require.NoError(t, err)
-
```
blampe, line 212: Worth calling out these tests introduced a race condition because we're now invoking methods on the user-provided config struct, and the tests are calling Configure in parallel. IMO it would be reasonable to error if the provider gets a second Configure call, but I'd like to handle that separately.

infer/tests/provider_test.go
blampe: As discussed this morning, the goal of allowing users to re-use existing components is still possible with the old API:

```go
func (r *RandomLogin) Construct(ctx *pulumi.Context, name string, typ string, args RandomLoginArgs, opts pulumi.ResourceOption) (*RandomLogin, error) {
    comp, err := NewRandomLogin(ctx, name, args, opts) /* existing code */
    if err != nil {
        return nil, err
    }
    return ctx.RegisterComponentResource(typ, name, comp, opts)
}
```

The PR includes `ComponentF` so users can still pass `NewRandomLogin` directly, but this prevents the user from benefiting from any of the framework's control over the schema (e.g., aliasing the component, deprecating it, etc.).

EronWright (approved):
A great usability improvement!

infer/tests/provider_test.go
iwahbe: I believe that you could still do all of that already by hanging the methods on the provider struct produced by the function.

infer/component.go
```diff
@@ -30,40 +30,72 @@ import (
...
+// To create an [InferredComponent], call the [ComponentF] function.
 type InferredComponent interface {
 	t.ComponentResource
 	schema.Resource
 
 	isInferredComponent()
 }
 
-func (derivedComponentController[R, I, O]) isInferredComponent() {}
+func (derivedComponentController[R, T, I, O]) isInferredComponent() {}
```
iwahbe, line 52: Why are there now 4 type arguments? They used to represent the resource holder, the inputs and the component resource.

infer/component.go
```diff
@@ -30,40 +30,72 @@ import (
...
+	return fn.construct(ctx, name, inputs, opts)
+}
+
+// ComponentF creates an [InferredComponent] using functions and types that an existing Pulumi component program
 // would have implemented.
 //
 // fn is the function you would use to construct the program.
 //
+// Note that this syntax doesn't allow annotating the component. Use
+// [infer.Component] for more control over the component's generated schema.
```
iwahbe, line 81: Before this PR, it did allow annotating the component.

infer/component.go
EronWright: True, and normally `derivedComponentController` extracts a type token plus annotations based on `R`.  However, in the special case of the functional style, it uses `O` because there is no `R`.  To support both modes, we need to abstract which type is reflected upon, hence `T`.

image

image

infer/component.go
EronWright: I agree this is untrue because annotations may be applied to the component resource; I'll remove this comment.

infer/tests/provider_test.go
EronWright: On @mikhailshilkov's point:
> The one explicit design goal of the Pulumi Components project was to keep the API of component definition of a "local" component and a "remote" component as close as possible, ideally exactly the same.

I have a concern that this goal was not achieved, if it is about the compatibility of a given component function as a local and remote component.  I say this because the user is now expected to call `p.GetTypeToken` within the function body, and this means that the function is effectively not a local component. Meanwhile, posing as a local component leaves you thread-bare when it comes to implementing advanced functionality, like a Call into the component.

mikhailshilkov: > the user is now expected to call p.GetTypeToken within the function body

This is not true. Our docs show it hard-coded, same as in other languages and local components.

infer/component.go
iwahbe: `Component` when took a function needed 2. `Component` when it took a struct needed 2. I don't understand why we would need to change it here. What do the 4 type arguments represent?

infer/tests/provider_test.go
EronWright: @mikhailshilkov thanks, I've been curious about this point, and will take steps to deprecate GetTypeToken.

infer/component.go
EronWright: I see three arguments, `Component[A,R,F]` before, and `ComponentF[A,O,F]` after.  R was renamed to O because R is typically the handler, not the component resource.

The fourth type argument is internal to `derivedComponentController` and simply serves to help it extract the type token from the appropriate source.

PR #368, opened 2025-05-07
Run provider with context

blampe:
This passes context to our Run methods so we can explicitly terminate providers.


examples/auto-naming/go.mod
```diff
@@ -9,7 +9,7 @@ toolchain go1.24.0
 require (
 	github.com/blang/semver v3.5.1+incompatible
 	github.com/pulumi/pulumi-go-provider v0.0.0-00010101000000-000000000000
-	github.com/pulumi/pulumi/sdk/v3 v3.165.0
+	github.com/pulumi/pulumi/sdk/v3 v3.168.1-0.20250507185715-bddbda9403ca
```
EronWright, line 12: Do we need to couple this to a pulumi upgrade?

EronWright (approved):

examples/auto-naming/go.mod
blampe: Assuming we cut an RC today, we should be able to point at a release by the time we cut 1.0.

PR #370, opened 2025-05-07
Add WithProvider and WithProviderF server options

blampe:
This consolidates all of the NewServer methods, and adds WithProvider and
WithProviderF options. This way servers can continue to be created with
concrete or lazy providers.

A `WithVersion` would make sense IMO, but we can do that later.

(It's odd that we use an option pattern here and a builder pattern for the provider.)

### DIFF
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -120,7 +120,7 @@ func NewServer(ctx context.Context, pkg string, version semver.Version,
 		opt.applyServerOption(o)
 	}
 	if o.mocks == nil {
-		o.mocks = &MockMonitor{} // Why?
+		o.mocks = &MockMonitor{} // MockResourceMonitor requires a non-nil monitor.
 	}
 	if o.provider == nil && o.providerF == nil {
 		return nil, fmt.Errorf("WithProvider or WithProviderF is required")
--- a/integration/integration.go
+++ b/integration/integration.go
@@ -26,7 +26,6 @@ import (
 	"testing"
 
 	"github.com/blang/semver"
-	provider "github.com/pulumi/pulumi-go-provider"
 	pprovider "github.com/pulumi/pulumi/pkg/v3/resource/provider"
 	presource "github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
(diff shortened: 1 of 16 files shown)


blampe:

- #368
    - #364
        - #370 ◀

integration/integration.go
```diff
@@ -61,49 +63,91 @@ type ServerOption interface {
...
 	opts ...ServerOption,
-) Server {
+) (Server, error) {
 	o := &serverOptions{}
 	for _, opt := range opts {
 		opt.applyServerOption(o)
 	}
 	if o.mocks == nil {
-		o.mocks = &MockMonitor{}
+		o.mocks = &MockMonitor{} // Why?
```
EronWright, line 124: Having a default mock instance allows the mock resource monitor that is created within newHost to assume there's a monitor to dispatch to.

integration/integration.go
```diff
@@ -61,49 +63,91 @@ type ServerOption interface {
...
 	host := newHost(ctx, o.mocks)
 
-	return &server{p.RunInfo{
+	var prov p.Provider
+	if o.provider != nil {
+		prov = *o.provider
+	}
+	if o.providerF != nil {
+		host.lazyInit()
+		hc, err := pprovider.NewHostClient(host.engineAddr)
```
EronWright, line 138: Inside `lazyInit` is a call to `NewEngineConn`, which is actually the same code as `NewHostClient`. In other words we are creating two connections but need only one.  Simply change the `host` struct to have `hostClient *pprovider.HostClient`, and then get use  `hostClient.EngineConn()` at will.

integration/integration.go
```diff
@@ -61,49 +63,91 @@ type ServerOption interface {
 	applyServerOption(*serverOptions)
 }
 
+// TODO: What does this do?
```
EronWright, line 66: See: https://github.com/pulumi/pulumi-go-provider/pull/363/files#diff-a3d303ad5b0de361788b41ef8df6b57b4ddb2fb21e841b4b2b1dbf2f67c83a3dR19

integration/integration.go
```diff
@@ -61,49 +63,91 @@ type ServerOption interface {
...
+func WithMocks(mocks pulumi.MockResourceMonitor) ServerOption {
+	return mocksOption{mocks: mocks}
+}
+
+// WithProvider backs the server with the given concrete provider.
+func WithProvider(p provider.Provider) ServerOption {
+	return providerOption{provider: p}
+}
+
+// WithProviderF backes the server with a lazily initialized provider.
```
EronWright, line 76:
```suggestion
// WithProviderF backs the server with a lazily initialized provider.
```

EronWright (approved):

integration/integration.go
blampe: Ahhhh thank you!

PR #371, opened 2025-05-07
Remove unnecessary modules

blampe:
These modules add unnecessary complexity without buying us anything. We can simplify things by removing them.


EronWright:
I personally like having each example be self-contained, e.g. I can copy it and start hacking. Also, the deps of the provider itself are likely different than the deps of an example, since the example might use various provider SDKs.

blampe:
> I personally like having each example be self-contained, e.g. I can copy it and start hacking.

Where do you copy it to? If you keep it in the examples directory that will continue to work. 

Currently, if you copy it anywhere outside the examples directory you need to manually remove the broken `replace`, then tidy. 

With this you can copy it anywhere inside the repo without needing to do anything. Or, if you copy it outside the repo, you would `go mod init pkgname` and then tidy.

So it's either less work or still two small manual steps, depending on where you copy it.

> Also, the deps of the provider itself are likely different than the deps of an example, since the example might use various provider SDKs.

That's OK. An example can use a different major version of a dependency, if that's what you mean. And remember consumers only pick up dependencies for the packages they use; in other words examples don't pollute the user's dependencies.

EronWright:
I may be convinced but it is not a blocker, right?

PR #373, opened 2025-05-08
Use property.Map with mock monitor

EronWright:
This PR is an ergonomic enhancement to the mock facility provided by the `integration` package, to use `property.Map` rather than `resource.PropertyMap`.  This makes it easier to make assertions about resource registration.

For example, a test case for the RandomLogin component can easily synthesize a random:RandomInteger resource.

```go
    NewResourceF: func(args integration.MockResourceArgs) (string, property.Map, error) {
        // validate the child resource inputs and synthesize their outputs
        case args.TypeToken == "random:index/randomInteger:RandomInteger" && args.Name == "login-length":
            assert.Equal(t, 8.0, args.Inputs.Get("min").AsNumber())
            assert.Equal(t, 24.0, args.Inputs.Get("max").AsNumber())
            return args.Name, property.NewMap(map[string]property.Value{
                "result": property.New(12.0),
            }), nil
```

This PR also adds a [new test case](https://github.com/pulumi/pulumi-go-provider/pull/373/files#diff-68c0e6c0cea0397afe13ee4565d0d89b27e55856228f9898ea24f71ea8b48844R189) to `examples/random-login/main_test.go`, demonstrating how to use mocks to test the `NewRandomLogin` function ([ref](https://github.com/pulumi/pulumi-go-provider/pull/373/files#diff-d83334518762585b7cfe144bd510474999dc319aaa71cdbfb553716a4b1ff988R100)).

I chose to simplify the RandomLogin and MoreRandomPassword components, with respect to the input and output property types. Resource references were being used, which I was unable to mock and I believe to be confusing to understand (as an example).

### DIFF
--- a/examples/random-login/main.go
+++ b/examples/random-login/main.go
@@ -48,21 +48,19 @@ func provider() p.Provider {
 	})
 }
 
-// TODO: Deserialization does not yet work for external resources. Right now, it looks
-// like this structure is only implementable in typescript, but that will need to change.
 type MoreRandomPasswordArgs struct {
-	Length *random.RandomInteger `pulumi:"length" provider:"type=random@v4.8.1:index/randomInteger:RandomInteger"`
+	Length pulumi.IntInput `pulumi:"length"`
 }
 
 type MoreRandomPassword struct {
 	pulumi.ResourceState
-	Length   *random.RandomInteger  `pulumi:"length" provider:"type=random@v4.8.1:index/randomInteger:RandomInteger"`
-	Password *random.RandomPassword `pulumi:"password" provider:"type=random@v4.8.1:index/randomPassword:RandomPassword"`
+	Length   pulumi.IntOutput    `pulumi:"length"`
+	Password pulumi.StringOutput `pulumi:"password"`
 }
 
 func NewMoreRandomPassword(ctx *pulumi.Context, name string, args *MoreRandomPasswordArgs, opts ...pulumi.ResourceOption) (*MoreRandomPassword, error) {
 	comp := &MoreRandomPassword{
-		Length: args.Length,
+		Length: args.Length.ToIntOutput(),
 	}
 	err := ctx.RegisterComponentResource(p.GetTypeToken(ctx), name, comp, opts...)
 	if err != nil {
@@ -70,7 +68,7 @@ func NewMoreRandomPassword(ctx *pulumi.Context, name string, args *MoreRandomPas
 	}
 
 	pArgs := &random.RandomPasswordArgs{
-		Length: args.Length.Result,
+		Length: args.Length,
 	}
 
 	config := infer.GetConfig[Config](ctx.Context())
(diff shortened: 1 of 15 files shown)


examples/random-login/main_test.go
```diff
@@ -185,3 +185,50 @@ func TestRandomSalt(t *testing.T) {
...
+		integration.WithMocks(&integration.MockResourceMonitor{
+			NewResourceF: func(args integration.MockResourceArgs) (string, property.Map, error) {
+				// mock the registration of the component's resources
+				switch {
+				case args.TypeToken == "random:index/randomId:RandomId" && args.Name == "login-id":
+					assert.Equal(t, 8.0, args.Inputs.Get("byteLength").AsNumber())
+					return "user", property.Map{}, nil
+
+				case args.TypeToken == "random:index/randomInteger:RandomInteger" && args.Name == "login-length":
+					assert.Equal(t, 8.0, args.Inputs.Get("min").AsNumber())
```
iwahbe, line 203: Nit (need not be addressed): I prefer not using `AsNumber()` in this scenario, since it can panic. This is equivalent:
```suggestion
					assert.Equal(t, property.New(8.0), args.Inputs.Get("min"))
```

integration/integration.go
```diff
@@ -120,7 +120,7 @@ func NewServer(ctx context.Context, pkg string, version semver.Version,
 		opt.applyServerOption(o)
 	}
 	if o.mocks == nil {
-		o.mocks = &MockMonitor{} // MockResourceMonitor requires a non-nil monitor.
+		o.mocks = &MockResourceMonitor{} // fake.MockResourceMonitor requires a non-nil monitor.
```
iwahbe, line 123: What's `fake` here?
```suggestion
		o.mocks = &MockResourceMonitor{} // MockResourceMonitor requires a non-nil monitor.
```

iwahbe (commented):
Using `property.Map` instead of `resource.PropertyMap` is great. 

I'm fine with removing the resource references from `random-login`, but I think thats the only example with resource references, and we should have at least one example with resource references. If you want to remove them here, can you add a new example that uses resource references?

examples/random-login/main.go
```diff
@@ -48,40 +48,39 @@ func provider() p.Provider {
 	})
 }
 
-// TODO: Deserialization does not yet work for external resources. Right now, it looks
-// like this structure is only implementable in typescript, but that will need to change.
```
blampe, line 52: Re: @iwahbe's comment, do external resources still only work in TS or is this outdated?

If they're only partially working I'm less concerned about keeping an example around.

iwahbe: There has been some work in pulumi/pulumi to support this, but I don't know its current extent.

examples/random-login/main_test.go
EronWright: Yeah that does look better. One little surprise upon trying this style is that it would assert over all components of the value, e.g. the deps.  For better/worse it led to:

```
        	Error:      	Not equal: 
        	            	expected: property.New(12.0)
        	            	actual  : property.New(12.0).WithDependencies([]urn.URN{"urn:pulumi:stack::project::random-login:index:RandomLogin$random:index/randomInteger:RandomInteger::login-length"})
```

examples/random-login/main.go
EronWright: OK, I adjusted the component-provider example to use a resource reference for the resultant RandomPassword.

blampe (approved):

PR #375, opened 2025-05-08
Use `ProviderBuilder` in the examples

EronWright:
This PR updates the examples to uniformly use the `ProviderBuilder` and the `provider.Run` function.

### DIFF
--- a/examples/assets/main.go
+++ b/examples/assets/main.go
@@ -11,6 +11,7 @@ import (
 	p "github.com/pulumi/pulumi-go-provider"
 	"github.com/pulumi/pulumi-go-provider/infer"
 	"github.com/pulumi/pulumi-go-provider/infer/types"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 )
 
 type HasAssets struct{}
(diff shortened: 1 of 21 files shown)


examples/component-provider/main.go
```diff
@@ -18,23 +18,33 @@ package main
...
+	"os"
 
+	p "github.com/pulumi/pulumi-go-provider"
 	"github.com/pulumi/pulumi-go-provider/examples/component-provider/nested"
 	"github.com/pulumi/pulumi-go-provider/infer"
 )
 
 func main() {
-	p, err := infer.NewProviderBuilder().
-		WithNamespace("example-namespace").
```
EronWright, line 28: @mikhailshilkov should these examples call `WithNamespace` and what would a good value be?

examples/random-login/main.go
```diff
@@ -20,32 +19,37 @@ import (
...
+func provider() (p.Provider, error) {
+	return infer.NewProviderBuilder().
+		WithResources(
+			infer.Resource[*RandomSalt](),
+		).
+		WithComponents(
 			infer.Component(NewRandomLogin),
 			infer.Component(NewMoreRandomPassword),
-		},
-		Config: infer.Config[Config](),
```
EronWright, line 37: @blampe should it be `infer.Config[Config]` or `infer.Config[*Config]`?

examples/component-provider/main.go
mikhailshilkov: I think it's a good practice, so that generated SDKs aren't namespaced to default `pulumi`. You don't like `example-namespace`?

examples/random-login/main.go
blampe: `infer.Config[Config]` is fine since it doesn't have any pointer methods.

blampe (approved):

examples/assets/main.go
```diff
@@ -19,17 +20,27 @@ type HasAssetsArgs struct {
 }
 
 func main() {
-	err := p.RunProvider(context.Background(), "assets", "0.1.0", provider())
+	provider, err := provider()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error: %s", err.Error())
+		os.Exit(1)
+	}
+	err = p.RunProvider(context.Background(), "assets", "0.1.0", provider)
```
EronWright, line 28: Will change to provider.Run()

examples/component-provider/main.go
EronWright: I don't think it very descriptive because the concept of namespaces lacks definition.  One might mistakenly think a namespace is at the provider level, to contain the provider's types. I believe it is at the org-level, to contain an organization's providers. 

How about `WithNamespace("examples")` since these examples are in an `examples/` folder.

EronWright:
Blocked on https://github.com/pulumi/pulumi-go-provider/pull/365

PR #377, opened 2025-05-09
Keep internal metadata under __internal

blampe:
The __internal key is already automatically excluded from diffs by the engine.


EronWright (approved):
LGTM

iwahbe (approved):

PR #378, opened 2025-05-10
Brackets for components

EronWright:


iwahbe:
When we announced that this as "the way" to build components, we said this syntax was locked. Lets not break people here.

blampe:
> When we announced that this as "the way" to build components, we said this syntax was locked. Lets not break people here.

@iwahbe @mikhailshilkov (re: https://github.com/pulumi/pulumi-go-provider/pull/365#discussion_r2084743637) I think there's confusion about what's actually changing with this + #365. Existing component closures aren't affected, we're just changing `infer.Component` to `infer.ComponentF` to make room for both styles. If we've said the closure syntax is locked that's fine and continues to be true, but go-provider is explicitly unstable.

Keep in mind our own examples look like:
```
type SecureBucket struct {
	pulumi.ResourceState

	BucketName pulumi.StringOutput `pulumi:"bucketName"`
}
```

So there's already a component struct involved. There are multiple ways to make the type known to the framework, and we're making it possible to share the struct in an alternative, more direct way _if the user needs that_. Eron has a ton of experience building components and I trust his judgment if he feels this is an important addition.

Between #378 and #365, I would say #378 is optional if we don't feel comfortable with the `Construct` API. `infer.ComponentF` can continue to support the existing call style, and `infer.Component` can always be added later.

mikhailshilkov:
> Between https://github.com/pulumi/pulumi-go-provider/pull/378 and https://github.com/pulumi/pulumi-go-provider/pull/365, I would say https://github.com/pulumi/pulumi-go-provider/pull/378 is optional if we don't feel comfortable with the Construct API. infer.ComponentF can continue to support the existing call style, and infer.Component can always be added later.

Sorry, is your proposal to make a breaking change from `Component` to `ComponentF` and be this the only change in this area?

blampe:
> But why? From my naive user's point of view, that's a breaking change to a worse name with no benefits for users.

The motivation is covered in #365 and you can see it in action in https://github.com/pulumi/pulumi-docker-build/pull/543. Being able to unit test your provider is an _enormous_ benefit for users.

EronWright:
I think that `ComponentF` is a better name in this situation, because it reduces confusion in a few situations.  For example, until v0.25 (March), `Component()` was the way to create a `ComponentResource`, which this PR brings back.  So if you're a user who hasn't ported to v1 yet, chances are you don't need to change anything.  

We accomplished the goal of allowing you to define a component using the (functional) "local component" syntax, but in my view, having a true ComponentResource is good too, and we've managed to solve for both.  With `ComponentF` we gain the additional benefit of being more compatible with code that is older than 0.25.

mikhailshilkov:
> I think that ComponentF is a better name in this situation, because it reduces confusion in a few situations. 

I think that `Component` is a better name because it doesn't have `F` in it, I guess most users won't be able to tell what `F` means here.

> For example, until v0.25 (March), Component() was the way to create a ComponentResource, which this PR brings back. So if you're a user who hasn't ported to v1 yet, chances are you don't need to change anything.

I'm confused now because

> I would say https://github.com/pulumi/pulumi-go-provider/pull/378 is optional if we don't feel comfortable with the Construct API. infer.ComponentF can continue to support the existing call style, and infer.Component can always be added later.

suggested we are not shipping `Component`, just `ComponentF`?

In any case, by default, I don't support having two ways to declare components because it's two ways to document, two ways to support, and it's very confusing for users that need to decide.

I apologise for jumping on a draft PR and commenting out of the blue, maybe I misunderstand something. If you do want to make a breaking change here, I request an RFC and a discussion with the Pulumi Components team.

EronWright:
How about we do this without a breaking change, because we are simply out of time. let's add the component resource variant using a different signature: ComponentResource()

blampe:
@mikhailshilkov the naming is consistent with the rest of the library. We have `RunServer` and `RunServerF`, for example. The latter takes a function for lazy binding.

It's more confusing for the user to pass structs everywhere _except_ for components:

```go
infer.Resource(&myResource{})
infer.Function(&myFunction{})
infer.Config(&myConfig{})
infer.Component(myComponentFunction) // ???
```

As Eron mentioned, it's not even obvious for us!

> I made a mistake a few times in writing test code, of writing infer.Component(RandomComponent) rather than infer.Component(NewRandomComponent), and then squinted at the compile error.

This usage is obvious and consistent with the F-suffix used here and in pu/pu:

```go
infer.Resource(&myResource{})
infer.Function(&myFunction{})
infer.Config(&myConfig{})
infer.Component(&myComponent{})
infer.ComponentF(myComponentFunction)
```

> In any case, by default, I don't support having two ways to declare components because it's two ways to document, two ways to support, and it's very confusing for users that need to decide.

I agree, and I wish more time had been spent on #313 to do this in a way more consistent with the rest of the library. (Take a type that implements an interface.) Unfortunately that ship sailed, but, again, the library is unstable so that we can experiment with and adjust these things. 

> I apologise for jumping on a draft PR and commenting out of the blue, maybe I misunderstand something. If you do want to make a breaking change here, I request an RFC and a discussion with the Pulumi Components team.

Please tell me what you would expect to see from an RFC that is not already captured in these PRs and discussions.

Worth mentioning the justification unchanged from #313:

> We believe that a combination of a preview 0.x version of the library, low number of end users for Components, and simplicity of migration warrants the breaking change to get clean API in the future.

PR #380, opened 2025-05-13
Infer from values instead of types (combined)

blampe:
This changes up our API to infer the provider's schema from, and to dispatch requests to, concrete instances instead of abstract types. For example, hooking up a resource now looks like `infer.Resource(&R{})` instead of `infer.Resource[*R]()`. It's a small syntactical difference but it gives the framework a lot more power and flexibility.

To round out this capability for both custom resources and component resources, this PR re-introduces the option of using a struct to define a component resource. This option was recently removed in https://github.com/pulumi/pulumi-go-provider/pull/313 in favor of a functional alternative that resembles that of a "local" component. With this PR, both styles are supported.

The first benefit is call signatures will feel more familiar to most Go developers. The reflect-ish `f[T]()` style makes types a very front-and-center concern while working, but our users just want to plug into some CRUD methods. We want something that gets out of the user's way, and simply passing something that implements an interface (`f(r)`) feels much more natural.

The second benefit is the user is now able to attach some "state" to their controllers (e.g. `&R{log: logger}`), which they can leverage during `Create`, `Update`, etc. This means it's now possible to unit test these providers with mock clients injected, whereas previously the only option available was integration testing. More specifically, users can now test provider logic using their schema's own Go structs without needing to dip into property maps. [Here](https://github.com/pulumi/pulumi-docker-build/pull/543/files#diff-39cd6456a9a1e4fd2f97f74f65ef2eeb5947907ba7e221121fba60c4e5c0c756L393-L413) is an example of an integration test modified to be a unit test instead.

As a corollary of invoking methods on the provided instances, it's also now easier to mock resources/components/functions. (This PR includes an [example](https://github.com/pulumi/pulumi-docker-build/pull/543/files#diff-39cd6456a9a1e4fd2f97f74f65ef2eeb5947907ba7e221121fba60c4e5c0c756R356-R366) where I use some mocks to confirm all of our methods are invoked on the user's instance.)

The last benefit that comes to mind is more for maintainers. Where previously we always had to go from { type → reflect → interface check }, there's now less of a need for the intermediate reflection layer. In other words, { instance → interface check } is now possible. Reducing the amount of reflection in the library is a big win for maintainability IMO.


examples/random-login/main_test.go
```diff
@@ -102,7 +102,7 @@ const schema = `{
       ],
       "aliases": [
         {
-          "type": "pkg:other:RandomLogin"
+          "type": "random-login:other:RandomLogin"
```
EronWright, line 105: Yes expected due to https://github.com/pulumi/pulumi-go-provider/pull/379

infer/tests/types_test.go
```diff
@@ -76,3 +78,135 @@ func TestOmittingAssetTypes(t *testing.T) {
...
+		r.EXPECT().Delete(gomock.Any(), gomock.Any())
+
+		prov, err := infer.NewProviderBuilder().WithResources(
+			infer.Resource(r),
+		).Build()
+		require.NoError(t, err)
+
+		err = prov.Delete(t.Context(), pgp.DeleteRequest{Urn: urn("TestReceivers", "delete")})
+		assert.NoError(t, err)
+	})
```
EronWright, line 211: Plz add a test for Component.

EronWright (approved):
This is a huge win for usability!

PR #384, opened 2025-05-16
New example: configurable

EronWright:
A simple demonstration of how the new "receiver" style in infer may be used effectively, e.g. to support mocking of an external system client.

This example also has tests based on the `integration` package for demonstration purposes, and uses gomock to mock the client.


blampe (approved):
Love it!

PR #395, opened 2025-06-03
Update link to the infer readme

mjeffryes:
A user pointed out that we were just linking to the top level API docs rather than anything specific to infer


iwahbe:
Ah, automerge. Sorry.

PR #400, opened 2025-06-06
adds mock invoke to fake.ResourceMonitorServer, adds unit test using …

jdavredbeard:
…invoke mocking, updates schema test, updates schema.json and sdk


examples/random-login/main.go
```diff
@@ -252,5 +269,29 @@ func (r *RandomSalt) WireDependencies(f infer.FieldSelector, args *RandomSaltArg
...
+
+type UsernameIsUnique struct{}
+
+func (UsernameIsUnique) Invoke(ctx context.Context, req infer.FunctionRequest[UsernameIsUniqueArgs]) (infer.FunctionResponse[UsernameIsUniqueRes], error) {
+	return infer.FunctionResponse[UsernameIsUniqueRes]{
+		// Here we would call the database to check if the username already exists.
+		// For now we will always return true, but we will mock different return values in the tests.
+		Output: UsernameIsUniqueRes{
+			IsUnique: true,
+		},
```
EronWright, line 283: Consider simplifying this function to return a boolean rather than `UsernameIsUniqueRes`.

examples/random-login/main_test.go
```diff
@@ -158,6 +187,13 @@ func TestSchema(t *testing.T) {
 	require.NoError(t, err)
 	blob := json.RawMessage{}
 	err = json.Unmarshal([]byte(s.Schema), &blob)
+	require.NoError(t, err)
+
+	var prettyJSON bytes.Buffer
+	err = json.Indent(&prettyJSON, blob, "", "  ")
+	require.NoError(t, err)
+	// t.Log(&prettyJSON)
```
EronWright, line 195: Remember to remove this stuff, seems temporary.

EronWright (commented):
This is looking great, the example is pretty fun.

EronWright (approved):

examples/random-login/main.go
jdavredbeard: I think I prefer it using the type rather than the boolean? For pattern's sake

EronWright:
As an aside, this hypothetical `usernameIsUnique` function would work best if the generated username are the same during preview and update.  But I don't think there's a way to do that yet.

Issue #292, opened 2025-02-08
Include the version of pulumi-go-provider in the provider state


mjeffryes:
We'd like to be able to learn about how provider authors are using pulumi-go-provider. Today we don't have any metrics that would tell us when providers are built on this library. The most basic metrics we could record would be to add the pulumi-go-provider version to the state of the provider resources created with this library so that we can see what versions of this library are in use by pulumi cloud users.

PR #408, opened 2025-08-22
Drop __internal usage

blampe:
Core doesn't want us abusing this field. We would still like the track the underlying information, though.

This PR lifts the relevant data to top-level properties prefixed with `__`. We also update diff logic to ignore these inputs along with the engine's "version" property, so we should not see any spurious updates related to these properties.

Refs https://github.com/pulumi/pulumi-docker-build/issues/579


guineveresaenger (commented):
Do we need to change what the README says, too? It says that "infer" is the "highest level" of this provider.

iwahbe:
This undoes https://github.com/pulumi/pulumi-go-provider/issues/292, right? Should we re-open the original issue?

blampe:
> This undoes #292, right? Should we re-open the original issue?

Maybe, I'll check with Matt when he's back in. I'm not sure if we're actually using this data.

mjeffryes:
We are using this and should probably move it to a new state field rather than just dropping it.

mcblair:
We upgraded to pulumi `v3.191.0` before upgrading our providers to `v1.1.1` and experienced the unexpected resource replacements. After upgrading our providers to `v1.1.1` and rolling it out, we fixed the unexpected resource replacements _but_ `refresh --preview --expect-no-changes` now always shows provider version changes and triggers a provider update. 

This is a deal breaker for us as we use that heavily for continuous drift detection. We've had to rollback and pin all pulumi versions to `v3.190.0` to keep drift detection functional.

PR #452, opened 2025-12-08
Don't set a default value for `"displayName"` or `"namespace`"

iwahbe:
Currently, we set a default value for both "displayName" and "namespace". Neither field needs a default value, and it is **highly** unlikely that the default values that we set (`"yourdisplayname"`, `"yournamespace"`) are useful to users.

The new logic for setting `"importBasePath"` (now divorced from `"yourdisplayname"` and `"yournamespace"`) is:

1. If the user set `"importBasePath"`, use that.
2. If the user set the repository, infer `"<repository>/sdk/go/<package-name>"`.
3. Otherwise, construct an import path only usable for local generation `"local-package/sdk/go/<package-name>"`.

### DIFF
--- a/examples/assets/schema.json
+++ b/examples/assets/schema.json
@@ -11,7 +11,7 @@
     },
     "go": {
       "generateResourceContainerTypes": true,
-      "importBasePath": "local-package/sdk/go/assets",
+      "importBasePath": "github.com/examples/assets/sdk/go/assets",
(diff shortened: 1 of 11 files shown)


iwahbe:
I've updated the logic so we keep `github.com/<namespace>/<pkg-name>/sdk/go/<pkg-name>` when `namespace` is set by the user.

PR #514, opened 2026-04-27
Resolve framework version via runtime/debug.ReadBuildInfo

iwahbe:
## Summary

- Replace the embedded `.version` file with a `Version()` function that resolves the framework's version at runtime from the consuming binary's Go build info (`runtime/debug.ReadBuildInfo`). Tagged releases report `v1.2.3`; untagged commits report a pseudo-version; framework self-builds report `(devel)`.
- Update the gRPC replay tests to call `Version()` instead of reading `.version`, so the expected and actual `__pulumi-go-provider-version` state values agree regardless of build mode.
- Delete `release.yml`. Its only remaining job was re-running lint and tests, which `main.yaml` already does on every push to `main`. With no `.version` file to keep in sync, there is nothing release-specific that has to happen at tag time — Go's module proxy picks up the new tag automatically.
- Switch the documented release process to direct GitHub release creation: maintainers click **Draft a new release** in the UI, which produces the tag and the release page together.

## Prior art

Other Go libraries already use this same `runtime/debug.ReadBuildInfo` + walk-`Deps` pattern to report their own module version at runtime:

- [slackhq/nebula `moduleVersion()`](https://github.com/slackhq/nebula/blob/2f4532f1026f78028c380aeb937d0f5baf41eab4/main.go#L293-L304) — reports the Nebula module version when embedded in another binary.
- [twmb/franz-go `softwareVersion()`](https://github.com/twmb/franz-go/blob/2a935ef6f0627e263490869ecca2321f93ca37a0/pkg/kadm/kadm.go#L93-L107) — supplies the kadm package version as the Kafka client software version.
- [tmc/langchaingo `UserAgent()`](https://github.com/tmc/langchaingo/blob/94f289e42739f8b0cf74f0a50f7342634fa3bdf9/httputil/useragent.go#L19-L46) — embeds the library version in the outgoing HTTP `User-Agent`.
- [buildkite/go-buildkite `Version`](https://github.com/buildkite/go-buildkite/blob/61840b79f1c68e8b72e816c8bc6d08ec3fb8b4bc/version.go#L9-L42) — falls back from `info.Main` to walking `info.Deps`, the same shape used here.
- [cosmos/cosmos-sdk `getSDKVersion()`](https://github.com/cosmos/cosmos-sdk/blob/ee19321c804df1d7bbf72476104c96f161d20288/version/version.go#L40-L57) — reports the linked Cosmos SDK version in `appd version` output.

## Test plan

- [x] `go test ./...` in the root module
- [x] `go test ./...` in `tests/`, including the gRPC replay tests that exercise `__pulumi-go-provider-version`
- [ ] Confirm on a tagged build that `Version()` returns the tag (will be exercised by the next release)


pgavlin (approved):
might be worth a little library for this

Frassle (approved):

PR #536, opened 2026-06-09
Apply the `modernize` lint

iwahbe:
This is another useful lint that LLMs tend to struggle to get right on their own.

### DIFF
--- a/infer/function.go
+++ b/infer/function.go
@@ -108,13 +108,13 @@ func fnToken(tk tokens.Type) tokens.Type {
 }
 
 func (r *derivedInvokeController[F, I, O]) GetSchema(reg schema.RegisterDerivativeType) (pschema.FunctionSpec, error) {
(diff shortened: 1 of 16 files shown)


infer/function.go
```diff
@@ -139,9 +139,6 @@ func objectSchema(t reflect.Type) (*pschema.ObjectTypeSpec, error) {
 	if err != nil {
 		return nil, fmt.Errorf("could not serialize input type %s: %w", t, err)
 	}
-	for n, p := range props {
-		props[n] = p
-	}
```
Frassle, line 144: 😆 what was this ever doing

Frassle (approved):

infer/function.go
iwahbe: Making sure `props` contains the contents of `props`? 🤷

PR #542, opened 2026-06-22
Expose `Handshake` services

iwahbe:


guineveresaenger (approved):
This looks like scaffolding so LGTM, but inquiring minds would love the context.

PR #574, opened 2026-08-24
Breaking change: Pass through full alias specs in ConstructRequest

### LINKED ISSUE #465: Pass through full alias specs
## Hello!

## Issue details

How aliases were passed was changed in https://github.com/pulumi/pulumi/pull/21050. https://github.com/pulumi/pulumi-go-provider/pull/464 patched over the issue so the SDK could take the update, but it didn't finish plumbing the change through, which we need to do.

iwahbe:
Fixes #465.

Since pulumi/pulumi#21050, aliases on a gRPC `ConstructRequest` arrive as either concrete URNs or alias specs (type, name, project, stack, parent, noParent). #464 patched the SDK update by keeping `ConstructRequest.Aliases` as `[]resource.URN`, which silently dropped every spec-form alias before it reached the component's nested resource registrations.

`ConstructRequest.Aliases` is now `[]resource.Alias`, and both translation directions (`newConstructRequest` and `ConstructRequest.rpc`) preserve the full alias, mirroring the conversions in pulumi/pulumi's `provider_server.go` / `provider_plugin.go`.

**Breaking change**: the type of `ConstructRequest.Aliases` changes from `[]resource.URN` to `[]resource.Alias`. A URN-only alias is represented as `resource.Alias{URN: ...}`.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -1384,7 +1384,7 @@ type ConstructRequest struct {
 	Inputs property.Map
 
 	// Aliases is the set of aliases for the component.
-	Aliases []presource.URN
+	Aliases []presource.Alias
 
 	// Dependencies is the list of resources this component depends on, i.e. the DependsOn resource option.
 	Dependencies []presource.URN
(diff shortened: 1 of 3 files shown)


julienp (approved):
Since we don't have a changelog, would it make sense to add `Breaking change: ...` to the title of the PR so it shows up in the release notes?

Issue #409, opened 2025-08-27
Reconsider default provider replacement behavior


blampe:
Currently any changes to a provider's config will trigger a replacement if it doesn't implement `DiffConfig`.

mjeffryes:
To elaborate a bit: We think this is probably not a good default since it means a change to the provider config will replace every resource created by that provider.

Since this would be a significant behavioral change, we intend to wait to change this until the next major version.


In the meantime, I would recommend that *all* users of this framework override `DiffConfig` with an implementation that does not replace unless you intentionally want to trigger resource replacements on provider config changes.

PR #580, opened 2026-08-25
Treat zero-valued value-typed optional fields as unset

### LINKED ISSUE #577: infer: `T,optional` diffs after import (`Read`)
> [!NOTE]
> This issue was authored by an AI agent on @iwahbe's behalf.

## What happened?

An `infer` provider declared its config with a value-typed optional field:

```go
type Config struct {
	ApiKey string `pulumi:"apiKey,optional" provider:"secret"`
}
```

With `apiKey` unset (the provider reads an environment variable in `Configure`
instead), `CheckConfig` encodes the receiver back into the checked inputs, so
the unset field comes back as an empty secret string. The stored provider
inputs (e.g. the bare `{version}` entry the CLI writes on `pulumi import`, or
the previous version's inputs on a provider upgrade) don't contain `apiKey`,
so `DiffConfig` sees the field as added — and since any non-`version` provider
config change is a replacement (#409), the provider is planned for replacement,
cascading delete-replace into every resource under it:

```
DiffConfig(...).oldInputs: version={0.0.1}
DiffConfig(...).newInputs: __pulumi-go-provider-infer={true}
                           __pulumi-go-provider-version={v1.6.0}
                           apiKey={[secret]}        # empty string, never set
                           version={0.0.2}
DiffConfig(...) success: changes=2 #replaces=[apiKey] diffs=#[version apiKey]
```

In practice (iwahbe/pulumi-resend v0.0.2, pulumi-go-provider v1.6.0): after
`pulumi import` of a `Domain`, the very next `pulumi preview` of the unchanged
program planned `+-resend:index:Domain (replace)` — a delete-then-create of the
production sending domain. Only `protect: true` stopped it. Every provider
version bump replayed the same cascade.

Changing the field to `ApiKey *string` fixes it: the nil pointer is omitted
from the checked inputs, and the provider diffs clean.

## What we'd like

`registerTypes` already rejects optional fields without pointer indirection for
structs and enums (`optionalNeedsPointerError`, added for #205). Please extend
that to **all** optional fields, primitives included: a value-typed optional
cannot distinguish "unset" from the zero value, and `Check`/`CheckConfig`
materializing the zero value is what turns this into a provider-replacement
footgun. Failing fast at schema/startup time — like #205 did for structs and
enums — would have surfaced the mistake immediately.

**Expected:** declaring `Field string `pulumi:"field,optional"`` (or any
non-nilable optional) is rejected at provider startup, instead of silently
producing checked inputs that diff against stored state.

## Environment

pulumi-go-provider v1.6.0; reproduced end-to-end in iwahbe/pulumi-resend
(regression test `TestUnsetApiKeyDoesNotReplaceProvider` replays the
CheckConfig → DiffConfig sequence and fails on the value-typed field).

iwahbe:
Currently, non-pointer optional fields cannot distinguish "unset" from zero, so `Check`/`CheckConfig` materialized the zero value into the checked inputs. Those materialized values then diffed against stored state (from `Read`, for example) that didn't go through `Check`/`CheckConfig`. For provider config, where any non-`version` change is a replacement (#409), an unset value-typed optional field like

```go
ApiKey string `pulumi:"apiKey,optional" provider:"secret"`
```

planned a provider replacement after a `pulumi import` or a provider upgrade.

## Semantics

Current semantics — a value-typed optional materializes its zero into the checked inputs:

| Input | Checked inputs | Type |
|-------|----------------|------|
| {}    | {f:0}          | int  |
| {f:0} | {f:0}          | int  |
| {f:4} | {f:4}          | int  |
| {}    | {}             | *int |
| {f:0} | {f:0}          | *int |
| {f:3} | {f:3}          | *int |

Desired semantics — Check is the identity on present keys and materializes nothing:

| Input | Checked inputs | Type |
|-------|----------------|------|
| {}    | {}             | int  |
| {f:0} | {f:0}          | int  |
| {f:4} | {f:4}          | int  |
| {}    | {}             | *int |
| {f:0} | {f:0}          | *int |
| {f:3} | {f:3}          | *int |

To avoid replaces on states written under the current semantics, diff obeys invarient:

```
Diff({f:zero(T)}, {}) == NoDiff  <=>  TypeOf(f) == T where !IsPointer(T)
```

**Zeros and absence are indistinguishable in diff exactly when the type can't distinguish them in Go.** A legacy state carrying a materialized `{f:0}` for a value-typed field compares equal to newly checked inputs that omit it — existing states never see a diff or a replacement from this change. For a pointer field, `new(0)` is a real value: equal values don't diff, and removing the field is a real change.

> [!WARNING]
> The biconditional above *requires* `Diff({f:0}, {}) == Diff` for pointer fields. A state written by a pre-fix release may carry a materialized `{f:0}` for a field that was value-typed **at the time**; if a later release migrates that field from `T` to `*T`, the legacy zero reads as a genuine `new(0)` and diffs. Migrating a field `T` → `*T` is only safe after at least one `pulumi up` on a release with the new semantics (Same steps persist the new checked inputs, draining materialized zeros from state). Provider authors should never ship the library upgrade and a `T` → `*T` migration in the same release. 

Users who need to distinguish `new(0)` from unset can use `*T`; with these semantics the stored representation of `T,optional` is identical to what `*T,optional` stores for the same program, so that migration is state-invisible (and non-diffing) for states written under the new semantics.

Fixes #577


infer/internal/ende/ende.go
```diff
@@ -388,32 +395,167 @@ func (e *ende) Encode(src any) (resource.PropertyMap, mapper.MappingError) {
...
+	path resource.PropertyPath, keep func(resource.PropertyPath) bool,
+) {
+	if typ == nil {
+		return
+	}
+	for typ.Kind() == reflect.Pointer {
+		typ = typ.Elem()
+	}
+	switch typ.Kind() {
+	case reflect.Struct:
```
julienp, line 453: What happens when v is secret or an output? Do we need to unwrap first and drop the zeros in there too?

infer/internal/ende/ende.go
```diff
@@ -118,6 +118,10 @@ type ende struct {
 	// secretDepth counts the secret wrappers enclosing the value currently being
 	// walked.
 	secretDepth int
+	// present holds the paths of optional value-typed primitive fields that were
+	// explicitly present in the decoded input, so Encode can distinguish a zero the
+	// user supplied (kept) from a zero the decode materialized (dropped).
+	present map[string]bool
```
julienp, line 124: For input -> checked inputs we have what goes in and can track this, but we also use ende for encoding outputs, where we can't track presence.

infer/internal/ende/ende.go
iwahbe: This is reflecting through the plain types the user has, not a property value. There are no secrets or outputs there (tracking issue: https://github.com/pulumi/pulumi-go-provider/issues/212).

infer/internal/ende/ende.go
iwahbe: For `Create` & `Update`, we use the inputs as the `encoder` source, so we track presence for the output fields that were also input fields.

julienp: Right, but if I have something that has an output `blah: 0` that doesn't have a corresponding input, won't we drop it now?

iwahbe: Yes.

I'd say that behavior is more correct, but its also technically breaking. We could ship this with a warning that provider authors who want to be able to set the zero value of their optional property need to switch the type from `T` to `*T`.

My guess is that most users who see the change will be pleased, and we can ship with a warning on the release notes.

julienp: Maybe a breaking change message in the changelog is enough?

julienp (approved):

PR #581, opened 2026-08-27
feat: let a component learn its own plugin identity

defangdevs:
## The gap

A component that registers a resource of **its own package** from inside `Construct` cannot tell the engine where that package's plugin comes from.

The registration goes straight to `ctx.RegisterResource`, which applies no package defaults. A registration made through a generated SDK does carry them — codegen bakes the version and the plugin download URL into `internal.PkgResourceDefaultOpts` — but a provider registering its own resource types from inside `Construct` never goes through that path.

The engine then records a default provider with both fields empty:

| Provider URN | `pluginDownloadURL` | Referenced by |
|---|---|---|
| `...::default_2_6_0_github_/api.github.com/<org>/<repo>` | present | 0 resources |
| `...::default` | **absent** | the self-registered resource |

The URL goes to the provider nobody needs; the one doing the work has none. The parent chain cannot supply a fallback either — the parent is a component, and a component records no provider of its own.

## Why it bites third-party providers only

With no URL recorded, the engine falls back to `github.com/pulumi/pulumi-<name>`. For a first-party package like `kubernetes` that is a real repository, so the omission is harmless and the pattern survives unnoticed. For a third-party package it is not:

```
could not create provider urn:...::pulumi:providers:my-pkg::default:
load plugin for my-pkg provider: could not find latest version for provider my-pkg:
404 HTTP error fetching plugin from
https://api.github.com/repos/pulumi/pulumi-my-pkg/releases/latest
```

It stays hidden for a long time, because two things have to miss at once:

- **`up` never fails.** Resolving the SDK-registered provider has already put the plugin in the local cache, and a versionless request matches any installed version of that plugin (`LegacySelectCompatiblePlugin`), so the bare provider finds it there and never calls out.
- **`destroy` does not run the program.** Nothing registers the SDK-side provider, so nothing populates the cache. On a machine that has the plugin installed anyway this still works; on a cold cache the bare provider has no address to fetch from and the stack cannot be destroyed at all. Recovery means hand-editing the checkpoint.

We hit exactly this on a stack that could not be torn down for four consecutive days.

## Why the values cannot come off the wire

`ConstructRequest` looked like the natural place to read them from, following what `pulumi-kubernetes`' `GetChildOptions` does — read `Version` and `PluginDownloadURL` off the incoming options and re-apply them to children. That does not work here: the Construct gRPC request has no field for either, so `newConstructRequest` has nothing to populate and the options handed to `Construct` never carry them. Only `Providers` survives the trip. (`pulumi-kubernetes` is unaffected because it is a hand-written provider on the Go SDK's `provider.Construct`, not on this framework.)

They can, however, come from the running provider, which knows both — the version already rides on `RunInfo`, and `infer` holds the download URL in its schema metadata. It just never shares them.

## The change

```go
func (c *Component) Construct(
	ctx *pulumi.Context, name, typ string, args Args, opts pulumi.ResourceOption,
) (*Component, error) {
	self := provider.GetPluginIdentity(ctx.Context())

	child := &ChildState{}
	err := ctx.RegisterResource("my-pkg:index:Child", name+"-child", args, child,
		append(self.ResourceOptions(), pulumi.Parent(comp))...)
	...
}
```

- `provider.GetPluginIdentity(ctx)` returns the `Package`, `Version` and `DownloadURL` of the plugin serving the request. Fields the provider never supplied come back empty; the call never fails.
- `PluginIdentity.ResourceOptions()` turns them into `pulumi.Version` / `pulumi.PluginDownloadURL`, skipping whichever is empty.
- `infer.Wrap` publishes the download URL on the context through the existing `mContext.Wrap` seam, next to the `RunInfo` the framework already puts there. Providers that set no URL are unaffected — no wrapper is added at all.

### Why a value and not a blanket option

The identity names one specific plugin, so it is returned to be applied per resource rather than forced onto every child of the Construct context. A component typically creates resources from other packages too, and handing them this provider's version and URL would send the engine looking for *their* plugin at *our* coordinates. Registering a stack transform would have the same problem, and a wider blast radius. The test covers both halves: the own-package child records the version and the URL, and a foreign-package child registered in the same Construct is left untouched.

This is also why the framework cannot simply do it for the author — only the call site knows which package a given `RegisterResource` belongs to.

## Test

`infer/tests/plugin_identity_test.go` reads `Version` and `PluginDownloadURL` off the `RegisterResource` RPC through `MockResourceArgs.RegisterRPC`. Removing the `infer` change makes it fail on the URL assertion. A second test covers a provider that set no download URL: the version still comes through and `ResourceOptions()` yields just the one option.

`make test_unit` and `golangci-lint run -c .golangci.yaml` are clean — the lint issues that remain (50 `goconst`, 2 `govet`) are pre-existing on `main` and untouched by this branch.

## Compatibility

Purely additive: one new exported type, one new function, one new internal context key. No existing signature or behaviour changes, and a provider that sets no `PluginDownloadURL` sees no difference.

### DIFF
--- a/infer/provider.go
+++ b/infer/provider.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 
 	p "github.com/pulumi/pulumi-go-provider"
+	"github.com/pulumi/pulumi-go-provider/internal/key"
 	t "github.com/pulumi/pulumi-go-provider/middleware"
 	"github.com/pulumi/pulumi-go-provider/middleware/cancel"
 	"github.com/pulumi/pulumi-go-provider/middleware/complexconfig" //nolint:staticcheck
@@ -170,6 +171,16 @@ func Wrap(provider p.Provider, opts Options) p.Provider {
 		provider.CheckConfig = config.checkConfig
 		provider = mContext.Wrap(provider, func(ctx context.Context) context.Context {
 			return context.WithValue(ctx, configKey, opts.Config)
+		})
+	}
+
+	// Publish the provider's own plugin download URL on the context, so a component
+	// registering a resource of this package from inside Construct can stamp it onto
+	// that resource. The Construct gRPC request carries no such field, and the version
+	// alongside it already rides on [p.RunInfo]. See [p.GetPluginIdentity].
+	if url := opts.PluginDownloadURL; url != "" {
+		provider = mContext.Wrap(provider, func(ctx context.Context) context.Context {
+			return context.WithValue(ctx, key.PluginDownloadURL, url)
 		})
 	}
 
--- a/infer/tests/plugin_identity_test.go
+++ b/infer/tests/plugin_identity_test.go
@@ -1,0 +1,200 @@
+// Copyright 2025, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"sync"
+	"testing"
+
+	"github.com/blang/semver"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	p "github.com/pulumi/pulumi-go-provider"
+	"github.com/pulumi/pulumi-go-provider/infer"
+	"github.com/pulumi/pulumi-go-provider/integration"
+	"github.com/pulumi/pulumi-go-provider/middleware/schema"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/pulumi/pulumi/sdk/v3/go/property"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+const identityDownloadURL = "github://api.github.com/example/pulumi-identity"
+
+// identityModuleMap maps this test package's Go package name onto the "index" module,
+// matching the rest of the suite.
+var identityModuleMap = map[tokens.ModuleName]tokens.ModuleName{"tests": "index"}
+
+// identityChild is a resource the component below registers itself, the way a component
+// provider registers a resource of its own package from inside Construct.
+type identityChild struct {
+	pulumi.ResourceState
+}
+
+type IdentityComponentArgs struct{}
+
+type IdentityComponent struct {
+	pulumi.ResourceState
+	IdentityComponentArgs
+}
+
+// NewIdentityComponent registers one child of its own package, stamped with the running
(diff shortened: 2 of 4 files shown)


iwahbe (commented):
Hi @defangdevs. Thanks for opening a PR. Unfortunately, doing this correctly is a little more complicated then exposing Version & PluginDownloadURL to a component. If the engine doesn't internally refer to the provider via the same PluginDownloadURL & Version the schema exposes, you can still get a mismatch.

The correct solution requires the engine to pass to the provider it's ID (Probably during `Handshake`), which the provider can then pass back to definitively say "serve this resource form me".

In the mean time, can you work around this by having your `PluginDownloadURL` hard-coded in your provider?

defangdevs:
Thanks — that makes sense, and the mismatch you describe is real: `(version, PluginDownloadURL)` is a proxy for identity, not identity itself, so a resource stamped this way still attaches to whichever default provider happens to share those coordinates rather than to the instance actually serving the component.

**On the workaround: yes, already done.** DefangLabs/pulumi-defang#479 hard-codes the URL and the version and threads them to the self-registered resource, and it is merged. We are unblocked, so nothing here is urgent for us.

**One observation on the engine plumbing**, in case it makes the real fix smaller than expected: the engine already sends a provider its own identity, and this framework already knows it is coming.

- `ConfigureRequest` carries `Urn`, `Name`, `Type` and `Id` — `sdk/go/common/resource/plugin/provider_plugin.go` builds it with all four (v3.256.0, around line 1122).
- The handshake flag that gates them is already surfaced here: `ConfigureWithUrn` in `provider.go`, documented as "the engine will send URN, Name, Type and ID to the provider as part of the configuration".
- But the adapter drops them. `(*provider).Configure` at `provider.go:917` builds the framework's `ConfigureRequest` from `Variables` and `Args` only, so nothing downstream can see the URN or the ID.

If that is right, a provider can already know its own reference without any protocol change; the missing pieces are this framework passing those four fields through, and a supported way to turn a `urn::id` reference into the provider option for a child registration. The Go SDK materialises exactly that shape today for `ConstructRequest.providers`, but through the unexported `newDependencyProviderResourceFromRef`. Doing it during `Handshake` instead, as you suggest, would also cover providers that are never configured — I have not checked whether Construct can be reached on that path.

I may well be missing why the Configure route does not work; you would know the constraints better than I do.

**How would you like to proceed?** I am happy to either rework this PR to pass the Configure identity through and register children against the provider's own reference, or close it and leave the interim accessor out of the framework — your call. If neither, closing is fine by me; the workaround covers our case.

lionello:
@iwahbe This PR was opened by my agent. I've been following it closely, but I don't know the internals of Pulumi well enough to judge whether this is correct. If you want it to rework this PR, you can nudge it by mentioning it: `@defangdevs`. Closing is fine too; we are unblocked.

iwahbe:
We are not going to take this as implemented. If you or @defangdevs want to open an issue describing the problem you are having, that would help ensure this gets seen later.

defangdevs:
Understood — closing this.

I opened #586 with the problem described on its own terms: what the component is trying to do, the two-default-provider checkpoint it produces, why `up` hides it and a cold-cache `destroy` does not, and why hard-coding the URL is a workaround rather than a fix. It also records your point that `(version, pluginDownloadURL)` is a proxy for identity rather than identity itself, and the `ConfigureRequest` observation from the comment above, in case it is useful when this gets picked up.

Thanks for the quick and clear review.

